PR #20, opened 2017-01-25
`resetPassword` promise returns a JSON response on success - fix #19

### LINKED ISSUE #19: Why only the `resetPassword` promise doesn't return a JSON response on success?
https://github.com/auth0/react-native-auth0/blob/master/authentication/api.js#L100

grifotv:


hzalaz:
Thanks for the PR but the endpoint does not return a valid JSON and it's only a string. A better change may be returning a dummy response, e.g. obj with email.

We'd doing some work on the lib so we might include this

PR #33, opened 2017-05-31
Guard against unavailable captureStackTrace

bigtiger:
`Error.captureStackTrace` was unavailable in our react-native app. It did not cause a compile error but, rather it failed at runtime with the error:

`Error.captureStackTrace is not a function. In 'Error.captureStackTrace(t,t.constructor)', 'Error.captureStackTrace' is undefined`

Node Version: 6.6.0

### DIFF
--- a/utils/error.js
+++ b/utils/error.js
@@ -1,7 +1,9 @@
 export default class Auth0Error extends Error {
   constructor (json) {
     super();
-    Error.captureStackTrace(this, this.constructor);
+    if (Error.captureStackTrace) {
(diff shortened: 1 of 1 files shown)


duhseekoh:
Yup... same issue... Lets merge this guy.

duhseekoh:
Hack to get around this until its merged. Throw this somewhere in your code.

```
if(!Error.captureStackTrace) { // captureStackTrace is only available when debugging
  Error.captureStackTrace = () => {};
}
```

PR #42, opened 2017-06-16
Update build tools

sebirdman:
later versions of gradle require > 25

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,7 +13,7 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion "25.0.3"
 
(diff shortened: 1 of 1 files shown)


hzalaz:
HI, thanks for the PR but I'd rather not update the build tools alone but also the SDK and tied dependencies like support libraries of android (like google custom tabs). 
Also since react-native-cli uses that by default I'd rather leave it to avoid confusing people that are not aware of the native platform (those who now can change it easily)

sebirdman:
@hzalaz that's fair. I set up my react-native project myself, so i'm using com.android.tools.build:gradle:2.3.0, maybe I should just downgrade that and call it a day instead.

hzalaz:
@sebirdman thanks, once I see that the cli has a new version I'll update. Will try try to research on how to make this easier to change

PR #59, opened 2017-07-06
Added audience to webAuth snippet

cocojoe:


cocojoe:
Added inline with native SDKs.

PR #70, opened 2017-08-02
Remove override annotation on createJSModules method

lukecwilliams:
React Native v0.47.0 removes the createJSModules call, so this @Override now breaks the build.

```
:react-native-code-push:compileReleaseJavaWithJavac - is not incremental (e.g. outputs have changed, no previous execution, etc.).
/Users/Luke/Projects/ClientPOC/node_modules/react-native-code-push/android/app/src/main/java/com/microsoft/codepush/react/CodePush.java:305: error: method does not override or implement a method from a supertype
    @Override
    ^
```

Confirmed this change fixes the build.  See RN Breaking Changes in the v0.47.0 release: https://github.com/facebook/react-native/releases/tag/v0.47.0

See also identical change e.g. in react-native-code-push: https://github.com/Microsoft/react-native-code-push/pull/895/commits/dff9f47a4b8b31e9ad4bfe36acf57fff8291163d

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Package.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Package.java
@@ -18,7 +18,7 @@ public class A0Auth0Package implements ReactPackage {
 
     // Required @Override for RN v0.46.x and earlier
     public List<Class<? extends JavaScriptModule>> createJSModules() {
-        return Collections.emptyList();
+      return Collections.emptyList();
     }
 
     @Override
--- a/android/src/main/java/com/auth0/react/A0Auth0Package.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Package.java
@@ -17,11 +17,6 @@ public class A0Auth0Package implements ReactPackage {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


hzalaz:
@lukecwilliams is this backwards compatible with those not in the bleeding edge of RN?

lukecwilliams:
I'm not 100% sure, though CodePush says their latest version is [only compatible with RN 0.47.0](https://github.com/Microsoft/react-native-code-push/releases/tag/v5.0.0-beta) and the pull request there said it's breaking, so I think the next version of react-native-auth0 would only be compatible with 0.47.0.

Side note - would you mind publishing new releases as Releases instead of Tags?  We and probably others are using tools that only track Releases!  Thanks a lot.

hzalaz:
@lukecwilliams try removing the override annotation. That will probably make this backwards compatible otherwise we need to think how to handle this (won't be the last since RN introduces breaking changes all the time)

lukecwilliams:
Removing @Override works for 0.47.0 and yes, should work in earlier versions.  Good idea.

Issue #87, opened 2017-08-30
Browser tabs not closed upon webauth success/failure


mleduque:
Each time I login, the WebAuth opens in a new browser tab.
After some roundtrip, I end up with a dozen login tabs in the browser.
How do I close the login tab in the browser ?

mleduque:
I this project dead ?

cocojoe:
If you need help then please provide sufficient information, what steps are required to recreate this issue starting with 
https://github.com/auth0-samples/auth0-react-native-sample/tree/master/00-Login

What platform/device are you testing with, what browser (presume this is Android) etc Let's see some screenshots or ideally a video.  

Thanks

cocojoe:
I've been testing and have noticed that browser tabs are not closed.  Thanks

lbalmaceda:
Adding the `Intent.FLAG_ACTIVITY_NO_HISTORY` to the intent might help, although it doesn't guarantee that the tab is going to be closed later.

mleduque:
@cocojoe I'll add more information soon, thanks.
@lbalmaceda I will try, thanks.

cocojoe:
@mleduque You can try PR #90 and feedback.

mleduque:
After Updating the dependency to 1.2.0, I can confirm it doesn't happen anymore.

PR #90, opened 2017-09-08
Close Browser app tab after finishing Auth

cocojoe:
Tested in Chrome 61
#87


cocojoe:
@lbalmaceda can you take a look at this first please.

lbalmaceda:
Seems fine. You should also try it without Chrome so it opens the link in a Browser instead of a Custom Tab. Btw, we don't have tests for this?

cocojoe:
I tried in browser, didn't work in the one I tested. RNA has tests but the platform specific stuff doesn't. Probably a bit of work to setup a CI that is testing JS/Android/iOS.

hzalaz:
I'd not do that to keep this simple. The native code should be minimal

PR #104, opened 2017-10-14
Updating README w/ info on iOS & Bundle Identifier

SaraChicaD:
I updated a section of the README that was giving me problems. I wanted to clarify what the `Bundle Identifier` should be in this instance, since I'm a Javascript developer not a native dev, and felt like more explicit instructions would have helped me, and my team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,14 @@ and then register a URL type entry using the value of `CFBundleIdentifier` as th
     </dict>
 </array>
 ```
+the end result should include _just_ the Bundle Identifier (ex: `samples.auth0.com`), and so look something like this:
 
+```xml
+<key>CFBundleURLSchemes</key>
+<array>
+    <string>samples.auth0.com</string>
+</array>
+```
 
 > The value `org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)` is the default for apps created with React Native CLI, you may have a different value.
 


README.md
```diff
@@ -118,7 +118,14 @@ and then register a URL type entry using the value of `CFBundleIdentifier` as th
     </dict>
 </array>
 ```
+the end result should include _just_ the Bundle Identifier (ex: `samples.auth0.com`), and so look something like this:
```
cocojoe: It might be clearer to incorporate something along the lines of 
-> the string value should be the literal value of the bundle identifier with no **$** variables

SaraChicaD: that works!! as someone who (prior to RN) never touched native code and is coming from web dev i think it's helpful for it to be super explicit, that was my intent here. thx.!

cocojoe: Agreed, if you can fix and commit and I'll merge it. Thanks

SaraChicaD: hey there it's not letting me edit this again, since i didn't make a branch or whatever, thanks.

cocojoe: I don't think I can do anything with it either now, so either fork it and make a new PR or I'll do it myself in a bit.

SaraChicaD: lol so silly, here's this: https://github.com/auth0/react-native-auth0/pull/106 thanks again!

cocojoe:
Moved to #106

PR #106, opened 2017-10-16
Updating README w/ info on iOS & Bundle Identifier (#2)

SaraChicaD:
I updated a section of the README that was giving me problems. I wanted to clarify what the `Bundle Identifier` should be in this instance, since I'm a Javascript developer not a native dev, and felt like more explicit instructions would have helped me, and my team. Updated to reflect comments from @cocojoe


README.md
```diff
@@ -119,6 +119,7 @@ and then register a URL type entry using the value of `CFBundleIdentifier` as th
 </array>
 ```
 
+The `<string>` value should just be the literal value of the Bundle Identifier with no $ variables, for example: `samples.auth0.com`.
```
cocojoe: Remove the word `just`, other than that, I'm happy.

Issue #124, opened 2018-01-22
react-native-auth0 WebAuth popup 


pragadeesh0106:
```
auth0
            .webAuth
            .authorize({ 
              scope: 'openid profile offline_access', 
              audience: 'https://' + Config.auth0Domain+ '/api/v2/'
              })
            .then(function (credentials) {
                console.log(credentials);
            })
            .catch(error =>
                console.log("error", error)
            );
```

when this function is called the webauth is opened outside the app . So when i go check my otp and click recent tabs in my phone my app remains there but the auth0 webauth popup is missing .

If anyone knows how to solve the issue kindly help...

cocojoe:
@pragadeesh0106 please can you tell us the steps to re-create, I presume you are using MFA.
It is expected behaviour that Web Auth is launched via your default system browser, most often Chrome.

pragadeesh0106:
web view opening in the browser is not a issue .

The auth0 page opens in a chrome web view and when i go check my otp  in my messenger and come back to enter otp im not able to reopen that webview it closese automatically.

scottbrassfield:
We are seeing this issue as well. When using webAuth with Passwordless authentication, the user is required to navigate out of the default system browser to retrieve the code from the entered email. Upon navigating back, the browser window where the code is meant to be entered has closed, and the authentication process cannot be completed.

cocojoe:
Thanks, I'll have a run through this.

msalokangas:
@cocojoe in our case, we tracked the problem to be related to Chrome Custom Tabs being used in Android. Right now our workaround will be to disable Custom Tabs.

pragadeesh0106:
@cocojoe what is the workaround to disable custom tabs does anyone know it ??

lbalmaceda:
@pragadeesh0106 I guess in [this](https://github.com/auth0/react-native-auth0/blob/master/android/src/main/java/com/auth0/react/A0Auth0Module.java#L56) method skip the "CustomTabsIntent" code and just start a new Intent with Action VIEW and the URL as URI, like it's done when the activity is null.

msalokangas:
@pragadeesh0106 @lbalmaceda yes, that's what we are doing. so we forked the library and removed the CustomTabsIntent part.

pragadeesh0106:
I didn't get you can u explain clearly what to do and i am calling the auth0 from react native so should i do the changes  in react native or the node modules of the react native auth0

msalokangas:
[...]

PR #126, opened 2018-01-29
Fix android web authentication

lbalmaceda:
I want to clarify how each flow works:

CustomTabs are meant to be open within the app. The NEW_TASK flag should not be used on that case. Of course I wasn't aware of this behavior until we required to pause the activity to get some data from an external app, like gmail.

For the Browser intent logic, which is only used in the case of calling `showUrl` from a non-activity class, the NEW_TASK flag is required as the context used is the Application one instead of the Activity one. If this flag is not provided a runtime exception raises. Browser is "still used" when going through the CustomTabs logic and no CustomTabs compatible browser app is installed in the device.

The NO_HISTORY flag introduced in a previous commit to avoid leaving the browser app in the "recents app" view doesn't work as expected. Instead, it's making the custom tab or browser session refresh the webpage on the activity resume. I've removed both flags and might be worth checking [this sdk](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/provider/CustomTabsController.java#L147) line as well.

This fixes https://github.com/auth0/react-native-auth0/issues/124

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.java
@@ -60,12 +60,11 @@ public class A0Auth0Module extends ReactContextBaseJavaModule implements Lifecyc
         if (activity != null) {
             CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder();
             CustomTabsIntent customTabsIntent = builder.build();
-            customTabsIntent.intent.addFlags(Intent.FLAG_ACTIVITY_NO_HISTORY);
-            customTabsIntent.intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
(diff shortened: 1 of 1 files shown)


cocojoe:
Have been testing on a Nexus 6 API 25 using Chrome. 

- HLP DB 👍 
- HLP Social / MFA 👍  
- Passwordless 👍 

Would be good to hear feedback from @msalokangas on this.

lbalmaceda:
I'll give them a few hours. Thanks @cocojoe

egorshulga:
After merging this PR the browser is being launched, I perform authentication, but then the browser is not being closed.
I have to press back button to return to the application.
I doubt this behavior was intended.

cocojoe:
@egorshulga Please open a new issue and add reproducible steps e.g. Testing setup, browser etc
Thanks

PR #131, opened 2018-02-07
Created definition file for ts support

ascariandrea:
Hi there, I've created definition files for `typescript`.
They're not super strict, but they can be enforced.


lbalmaceda:
@ascariandrea Hi! Thanks for the contribution. We don't plan to support definitions yet but you can create a PR in [this repo](https://github.com/DefinitelyTyped/DefinitelyTyped) and add support for this library that way. Cheers

PR #136, opened 2018-03-07
Fixed broken podspec for Cocoapods installations

danieljvdm:
The current podspec is invalid and will cause errors in `pod install`. This update fixes these issues and adds better version control that is tied to the npm version.


LcTwisk:
Can we merge this 😋?

lbalmaceda:
@cocojoe can you review this one please?

cocojoe:
I don't believe it was intended to be used as a pod, it's not something listed in the installation. It's never been published, just looks like a place holder podfile.
The method of installation is via npm.

What are you doing with it?

LcTwisk:
@cocojoe What is the reason of having a Podfile if there is no intention to use it with Cocoapods 😜? Most RN libraries support Cocoapods nowadays (even through `react-native link`). Probably because linking static libraries can result in conflicting dependencies while Cocoapods is able to resolve them.

Btw, npm and pods go perfectly together. This is how I'm using the Auth0 library with the proposed changes in this PR:
`pod 'A0Auth0', :path => '../node_modules/react-native-auth0'`

cocojoe:
If this is just used internally then maybe yes, no harm.

danieljvdm:
@cocojoe you don't have to publish for it to work on Cocoapods. Cocoapods is the industry standard in iOS development, and most people who are working on serious, large projects are going to be using Cocoapods. It's a pretty serious limitation not to support it. At the very least, you can support it unofficially by just merging this PR. There already is a podspec that is completely broken, why not fix it?

danieljvdm:
Furthermore, I couldn't even get this library to work _without_ having to fork this repo. If you're running an iOS setup with React Native installed via Cocoapods, the search headers will be completely wrong and this project won't be able to find and use the React framework.

danieljvdm:
@cocojoe thanks a bunch!

PR #141, opened 2018-03-26
fix wrong error format

finkef:
`webauth##authorize` does not parse errors returned from the API correctly which results in `Error: unknown error` messages. Instead, whenever the `error_description` field is available, it should use `AuthError` rather than `Auth0Error`.


lbalmaceda:
Can you please provide the original error responses from the server that were not correctly parsed?

cocojoe:
Closing inactivity

Issue #67, opened 2017-07-25
Could not invoke A0Auth.showUrl


VERSAYANA:
I followed the quick start for react-native.
Now I'm trying to show the login page with
`		auth0.webAuth
			.authorize({
				scope: 'openid email',
				audience: 'https://ramtin.auth0.com/userinfo'
			})
			.then(credentials => console.log(credentials))
			.catch(error => console.log(error));
`

But I'm getting "Could not invoke A0Auth.showUrl" error

"react-native": "0.44.0",
"react-native-auth0": "^1.0.3",

jaga-me:
having the same issue

hzalaz:
@jaga-me @VERSAYANA did you run react-native link react-native-auth0 as detailed in the README?

VERSAYANA:
@hzalaz Yes

ronitslyper:
Same issue, working fine on IOS simulator, but getting this error on Android simulator. How can I help debug?

knyuwork:
Same here.

hzalaz:
Is this reproducible with our samples?

ronitslyper:
@hzalaz can you point me at the sample to run?

Some more data. From the android debug monitor, the error is this:
![screen shot 2017-07-31 at 4 33 00 pm](https://user-images.githubusercontent.com/7136941/28782184-194ddafe-7615-11e7-848e-e8e469916bcb.png)


Did the usual clean, invalidate caches, build, clean, etc.., and changed (oddly) to this:
![screen shot 2017-07-31 at 4 42 23 pm](https://user-images.githubusercontent.com/7136941/28782034-b3e5fb24-7614-11e7-8c41-8d76a2a69249.png)

It does seem to be some kind of linking / build tools version error, but I don't see anything out of place. I'm using:
  compileSdkVersion 23
    buildToolsVersion "23.0.1"
with
compile "com.android.support:appcompat-v7:23.0.1"
    compile "com.facebook.react:react-native:+"

hzalaz:
You can download from https://auth0.com/docs/quickstart/native/react-native

ronitslyper:
The sample works fine on the same simulator.

ronitslyper:
Tried to make all the versions of packages / tools the same in the sample as in my app - still no crash in the simulator. @hzalaz , any ideas on how to proceed?

michaelspeed:
yes getting the same issue

hzalaz:
[...]

PR #149, opened 2018-04-22
android: allow root project to specify dependency versions

mlc:
React Native projects with multiple native dependencies often end up unintentionally relying on a variety of Android SDK versions, which can lead to unpredictable build-time failures or even run-time crashes, such as the weirdness in #67. Many libraries (see, e.g., react-community/react-native-maps#2047, invertase/react-native-firebase#1007) have been moving to a system of allowing the solution [recommended by google](https://developer.android.com/studio/build/gradle-tips.html#configure-project-wide-properties) of having the root project set properties with the desired dependency versions. This PR implements that solution for `react-native-auth0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ android:windowSoftInputMode="adjustResize">
 </activity>
 ```
 
+In order to reduce the potential for conflict which may be caused by the presence of differing versions of the Android support libraries, you may wish to [configure project-wide properties](https://developer.android.com/studio/build/gradle-tips.html#configure-project-wide-properties) setting the compile SDK version, build tools version, target SDK, and support library, in your root project's `build.gradle`, which will then be respected by `react-native-auth0` and many other React Native modules.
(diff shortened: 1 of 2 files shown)


demoran23:
This did the trick for me, after setting supportLibVersion in my root build.gradle's ext block.

lbalmaceda:
> after setting supportLibVersion in my root build.gradle's ext block.

@mlc @demoran23 should we add this notice somewhere in the Readme file??

mlc:
Good idea, @lbalmaceda! I've updated the PR to include a note in the `README`.

Amnesthesia:
I'm stuck on this but have copied the gradle file from this PR — any update on when this will be published on NPM?

lbalmaceda:
@Amnesthesia Sorry about that 😞will be in `1.3.0` soon when that release PR get's approved ⚡️

PR #158, opened 2018-06-05
Use SFAuthenticationSession on iOS11+

matt-oakes:
https://developer.apple.com/documentation/safariservices/sfauthenticationsession

Shares the cookies from Safari which allows a smoother sign-in experience if the user is already signed in on Safari.

Prompts the user to allow access before opening the Safari View Controller.

This API will be deprecated in iOS12 and replaced with [ASWebAuthenticationSession](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession). I think it's good to put this in for now though until the iOS12 SDK comes out of beta. As the API is basically the same as well, it will only require a small change to the new `showAuthorization` method to support both.


cocojoe:
Thanks @matt-oakes for the contribution, I'll need to check this out and have a play.
Not had a chance to play with `ASWebAuthenticationSession` yet, does it get rid of the aggressive dialog prompt?

Can you rebase please as well.

lbalmaceda:
@cocojoe can you please review this one?

cocojoe:
@matt-oakes sorry for the long delay on this, Why doesn't it by default switch to use `showAuthorization` for iOS 11+. It appears to be an opt-in option. I don't mind the idea of being able to opt-out for whatever reason. But it should be default behaviour.

Also as `ASWebAuthenticationSession` is now available for iOS12 and deprecates `SFAuthenticationSession` it would be a good time to add in a code path for iOS 12+ to switch to that.

Thanks

PR #176, opened 2018-09-18
Update Minimum iOS Deployment Target, inline with React Libs

coosamatt:
Updated Xcode build version from 7.0 to 9.0 to build for iOS 12.


cocojoe:
Thanks for the PR. Where does it say 9.0 is the minimum? You can still select 8.0 in Xcode 10.  I mean I think 9.0 or even 10.0 is a sensible minimum these days. 
I was looking for the minimum target that react native itself recommends (or defaults to)?
Our other iOS libs use 9.0 so in general this is fair.

cocojoe:
I'll need to test this, will set some time aside tomorrow and run the whole thing through latest RN.

coosamatt:
Updated to 8.0, I did find that works as expected.

cocojoe:
Had a look at this, so I saw the *Warning*. Having taken a look at the other Libs, given that all of the React* libs have a minimum deployment target of 9.0. It makes sense for us to follow this.
@coosamatt If you can set back to 9.0 thanks.

ios/A0Auth0.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cocojoe: Don't add this file.

coosamatt:
@cocojoe updated to 9.0 and removed IDE file.

cocojoe:
Fixes #179

PR #177, opened 2018-09-25
Allow authorize options to override default values

kenzic:


cocojoe:
@kenzic please can you provide more information in your description, what issue did you run into that required this PR.

cocojoe:
@kenzic Yes, why do you need to do this? The callback URL is standardised across our native platforms.

kenzic:
@cocojoe I'm running into issues using this along side a framework I'm using. But that's more of an aside. It seems strange that the defaults override the options. Is this intentional?

kenzic:
@cocojoe following up on this

cocojoe:
@kenzic well the previous comment still stands. https://github.com/auth0/react-native-auth0/pull/177#issuecomment-428876227

This is not something you can change (intentional) in the dedicated iOS or Android Auth0 libraries, so I would like to know what is the use case here for changing the callback URL.  What is your scenario, what framework etc Thx

kenzic:
The framework I'm using is Expo. Due to a bug in Expo I can't have a custom scheme URL and have the bundle identifier handle linking properly.

For example, if I have a bundle identifier `com.test.myapp` and a scheme `myapp` only one will work. The other will have it's protocol changed to exp. I'm working with the expo team to address this.

Right now, if my app supports the custom [URL scheme](https://developer.apple.com/documentation/uikit/core_app/allowing_apps_and_websites_to_link_to_your_content/defining_a_custom_url_scheme_for_your_app), the react-native-auth fails. All this being said, it does seem like there are other reasons someone might want to have a custom redirectUri.

My specific issue aside, the reason for my request comes down to:
- Should defaults override options?
- And/or should the API allow for changing the redirectUri?

defaults overriding options is seems like either a bug or a bit of a misnomer, and being able to modify the redirectUri seems like it would be pretty useful.

cocojoe:
Thanks for this answer, it's much clearer.

So I agree with you that user applied options should override defaults, otherwise there is not much value in the options.
I was hesitant around the callback. However, I can't see a good story around not having flexibility around this.

So please rebase your PR against latest master and add some tests. 👍

kenzic:
@cocojoe - rebased and pushed

cocojoe:
@kenzic This could be an issue with anyone who has set weird options but as these would never have been set due to the parameter order merging it would have worked okay. Is a bit of a weird scenario but wanted to voice it.

webauth/index.js
```diff
@@ -54,12 +54,12 @@ export default class WebAuth {
       }/${bundleIdentifier}/callback`;
       const expectedState = options.state || state;
       let query = {
-        ...options,
+        ...defaults
```
cocojoe: That missing comma!!

PR #182, opened 2018-10-11
Added id_token validation during code exchange

luisrudge:
### Changes

- Added id_token validation during code exchange.
- Because you need a `nonce` to validate the id_token, I added a new param to the `webAuth .exchange` method: https://github.com/auth0/react-native-auth0/pull/182/files#diff-7ebf6abd2439364f1fcc1ea4d1f20d5fR123
- When this param is provided, this library will validate the received id_token and throw an error if it's invalid (`{error: 'invalid_token'}).
- By default, all calls to `webAuth.authorize()` will send a nonce and validate the id_token accordingly

### Testing

* [x] This change has been tested on the latest version of the platform/language

### DIFF
--- a/webauth/index.js
+++ b/webauth/index.js
@@ -86,7 +86,7 @@ export default class WebAuth {
             status: 0
           });
         }
-        return client.exchange({ code, verifier, redirectUri }, 'foo');
+        return client.exchange({ code, verifier, redirectUri }, options.nonce);
       });
     });
   }
--- a/webauth/index.js
+++ b/webauth/index.js
@@ -86,7 +86,7 @@ export default class WebAuth {
             status: 0
           });
         }
-        return client.exchange({ code, verifier, redirectUri });
+        return client.exchange({ code, verifier, redirectUri }, 'foo');
       });
     });
   }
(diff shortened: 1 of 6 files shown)


cocojoe:
@luisrudge please add a decent PR description, something like https://github.com/auth0/lock/pull/1524/files 😄 would be a good start.

luisrudge:
@cocojoe done

cocojoe:
@luisrudge to be clear an autogenerated `nonce` is added by default?

luisrudge:
@cocojoe I don't think it is 🤔

cocojoe:
@luisrudge this make it sound like it is `By default, all calls to webAuth.authorize() will send a nonce and validate the id_token accordingly`

webauth/index.js
```diff
@@ -86,7 +86,7 @@ export default class WebAuth {
             status: 0
           });
         }
-        return client.exchange({ code, verifier, redirectUri });
+        return client.exchange({ code, verifier, redirectUri }, options.nonce);
```
esarafianou: Should the user of this library add the options.nonce?
e.g.
```
auth0
 .webAuth
 .authorize({scope: 'openid email', audience: 'https://{AUTH0_DOMAIN}/userinfo'}, nonce:<RANDOM_STRING>)
```
if so, how does [this](https://github.com/auth0/react-native-auth0/blob/eab2e1d6b3ad638fe772013cfdd63a280469594e/auth/__tests__/index.spec.js#L65-L68 ) test pass? 

If that's not the case and the nonce is not passed as an option by the user, how does it get a value?

luisrudge: This test has nothing to do with the nonce.. it's only validating the constructor call.

I'm not really sure why I'm using `options.nonce` instead of something auto generated. I'll have to dig through the PR again to test it.

esarafianou: Is there any update on that one? I'll then sign off from security.

PR #187, opened 2018-11-07
Added iOS SF/ASWeb AuthenticationSession support

cocojoe:
Adds support for iOS 11 SFAuthenticationSession & iOS 12 ASWebAuthenticationSession
This enables SSO in iOS 11+

No changes are required by the developer, all is handled internally.

### DIFF
--- a/ios/A0Auth0.m
+++ b/ios/A0Auth0.m
@@ -37,13 +37,13 @@ RCT_EXPORT_METHOD(hide) {
 }
 
 RCT_EXPORT_METHOD(showUrl:(NSString *)urlString closeOnLoad:(BOOL)closeOnLoad callback:(RCTResponseSenderBlock)callback) {
-    if (@available(iOS 11.0, *)) {
+    self.sessionCallback = callback;
+    self.closeOnLoad = closeOnLoad;
+    if (@available(iOS 13.0, *)) {
         [self presentAuthenticationSession:[NSURL URLWithString:urlString]];
     } else {
         [self presentSafariWithURL:[NSURL URLWithString:urlString]];
     }
-    self.sessionCallback = callback;
-    self.closeOnLoad = closeOnLoad;
 }
 
 RCT_EXPORT_METHOD(oauthParameters:(RCTResponseSenderBlock)callback) {
@@ -89,12 +89,12 @@ RCT_EXPORT_METHOD(oauthParameters:(RCTResponseSenderBlock)callback) {
                                           if ([[error domain] isEqualToString:ASWebAuthenticationSessionErrorDomain]) {
                                               switch ([error code]) {
                                                   case SFAuthenticationErrorCanceledLogin:
-                                                      callback(@[ERROR_CANCELLED]);
+                                                      callback(@[ERROR_CANCELLED, [NSNull null]]);
                                               }
                                           } else if(error) {
-                                              callback(@[error]);
+                                              callback(@[error, [NSNull null]]);
                                           } else if(callbackURL) {
-                                              callback(@[@{@"url": callbackURL.absoluteString}]);
+                                              callback(@[[NSNull null], callbackURL.absoluteString]);
                                           }
                                           self.authenticationSession = nil;
                                       }];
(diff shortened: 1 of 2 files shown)


webauth/agent.js
```diff
@@ -17,10 +17,12 @@ export default class Agent {
         resolve(event.url);
       };
       Linking.addEventListener('url', urlHandler);
-      NativeModules.A0Auth0.showUrl(url, closeOnLoad, err => {
+      NativeModules.A0Auth0.showUrl(url, closeOnLoad, result => {
         Linking.removeEventListener('url', urlHandler);
-        if (err) {
-          reject(err);
+        if (result && result.url) {
```
joshcanhelp: This seems odd ... if there is a `result`, it could be a success or an error? Is there maybe an `error_code` property on there when it's an error? Just seems like a non-standard way to pass a result/response to a callback.

webauth/index.js
```diff
@@ -54,7 +54,7 @@ export default class WebAuth {
       }/${bundleIdentifier}/callback`;
       const expectedState = options.state || state;
       let query = {
-        ...defaults
```
joshcanhelp: Was this even working before? Seems like a syntax error:

https://github.com/auth0/react-native-auth0/blob/master/webauth/index.js#L57

cocojoe: No, was in a previous previous 3d party PR. This is a fix for that.

webauth/agent.js
cocojoe: So I thought about this at the time, I could change the result to have a status e.g. success or error. Then you're changing the response to the end user as some errors have just been passed straight out from the Native Platform iOS/Android into err. Although looking at certainly iOS errors I don't think the JS side is equipped to process them anyway.

Perhaps we should more strongly define the errors from the native platforms or we need to change the callback method.
Currently using `RCTResponseSenderBlock` which only accept 1 argument which is an array.
https://facebook.github.io/react-native/docs/native-modules-ios

It might be better to use `RCTPromiseResolveBlock` although this becomes more of a refactor in both JS/iOS/Android to achieve this.

taboulot:
@cocojoe 
Any update on when your PR will be merged 😃

cocojoe:
I need to make some changes as I'm not happy with the implementation, even though it works 😄 
I should hopefully have some time later this week to pick this up again, thanks.

MSSPL-PiyaleeMaiti:
I am facing the same issue when I can get this updated version.

cocojoe:
I did pick this back up and got delayed, I am on it this week.

cocojoe:
@taboulot @MSSPL-PiyaleeMaiti If you are up for trying out this branch and feeding back I would appreciate it.

taboulot:
@cocojoe We will probably test it next will 👍Thanks !

ios/A0Auth0.m
```diff
@@ -10,8 +13,12 @@
 #import <React/RCTUtils.h>
 #endif
 
+#define ERROR_CANCELLED @{@"error": @"a0.session.user_cancelled",@"error_description": @"User cancelled the Auth"}
+#define ERROR_FAILED_TO_LOAD @{@"error": @"a0.session.failed_load",@"error_description": @"Failed to load url"}
```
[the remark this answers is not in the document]
cocojoe: These errors are from moving existing error responses to constants. So there is no actual change of code, only location of code.

ios/A0Auth0.m
```diff
@@ -10,8 +13,12 @@
 #import <React/RCTUtils.h>
 #endif
 
+#define ERROR_CANCELLED @{@"error": @"a0.session.user_cancelled",@"error_description": @"User cancelled the Auth"}
```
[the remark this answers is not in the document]
cocojoe: Same as before

taboulot:
@cocojoe 

We tested it and it works perfectly 👍

cocojoe:
Thanks, will get this released soon.

expatyler:
Any word on the release date for this @cocojoe  ?

cocojoe:
Sorry for delays, released 👍

Issue #191, opened 2018-11-29
Installing via CocoaPods hangs on installing obsolete React 0.11.0


rnevius:
### Description

When attempting to install via CocoaPods via `pod install` the process hangs and fails at `Installing React (0.11.0)`. The `React` pod [is deprecated](https://cocoapods.org/pods/React) and, since React Native already requires React, we should remove this dependency from `A0Auth0.podspec`.

#136 "fixes" the podspec while simultaneously introducing this bug, for recent versions of React Native.

### Prerequisites

* [x] Did you check the documentation?
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository?
* [x] Are there any related or duplicate Issues or PRs for this issue? 

### Environment

- React Native Auth0 version 1.3.1
- React Native version 0.56.0
- iOS Setup using CocoaPods, as required by the Firebase module for React Native

### Reproduction

With Auth0 for React Native installed, simply attempt to `pod install` from the `ios/` directory in an ejected app that uses CocoaPods. 

```bash
$ pod install
Analyzing dependencies
Fetching podspec for `A0Auth0` from `../node_modules/react-native-auth0`
Downloading dependencies
Installing A0Auth0 (1.3.1)
Using Firebase (5.11.0)
Using FirebaseAnalytics (5.3.0)
Using FirebaseAnalyticsInterop (1.1.0)
Using FirebaseCore (5.1.6)
Using FirebaseInstanceID (3.3.0)
Using FirebaseMessaging (3.2.1)
Using GoogleAppMeasurement (5.3.0)
Using GoogleUtilities (5.3.6)
Using Protobuf (3.6.1)
Installing React (0.11.0)  <-- hangs/fails
```

lbalmaceda:
Solved in PR 192.

PR #192, opened 2018-11-29
Remove obsolete React dependency from podspec.

rnevius:
### Changes

The React pod is deprecated and, since React Native already requires React, we should remove this dependency from A0Auth0.podspec.

### References

This resolves issue #191 .

### Testing

With Auth0 for React Native installed, simply attempt to `pod install` from the `ios/` directory in an ejected app that uses CocoaPods. This pod and all other pods should successfully install.

* [x] This change has been tested on the latest version of the platform

### Checklist


cocojoe:
Thanks for the PR, I read your issue and was going to suggest you raise a PR as you had the answers 😄

PR #203, opened 2019-01-18
Add Login with Client Credentials Grant

rafaelmaeuer:
### Changes

- method added to enable login using the Client Credentials Grant (M2M flow)
- necessary for using Auth0 on a react native app without user handling
- used for generic protection of API communication

### Testing

Implementation example for testing added to readme

* [X] This change adds unit test coverage
* [X] This change has been tested on the latest version of the platform/language or why not

### Checklist

* [X] All existing and new tests complete without errors
* [ ] All active GitHub checks have passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ In the file `android/app/src/main/AndroidManifest.xml` you must make sure the **
 </intent-filter>
 ```
 
-So if you have `samples.auth0.com` as your Auth0 domain you would have the following **MainActivity**  configuration:
+So if you have `samples.auth0.com` as your Auth0 domain you would have the following **MainActivity** configuration:
 
 ```xml
 <activity
@@ -78,7 +78,6 @@ android:windowSoftInputMode="adjustResize">
 ```
 
 > For more info please read [react native docs](https://facebook.github.io/react-native/docs/linking.html)
-
 
 In order to reduce the potential for conflict which may be caused by the presence of differing versions of the Android support libraries, you may wish to [configure project-wide properties](https://developer.android.com/studio/build/gradle-tips.html#configure-project-wide-properties). Setting the Compile and Target SDK versions, Build Tools version, and Support Library version in your **root project's** `build.gradle` file will make `react-native-auth0` and many other React Native modules use them.
 
@@ -134,7 +133,6 @@ Callback URLs are the URLs that Auth0 invokes after the authentication process. 
 
 > Callback URLs must have a valid scheme value as defined by the [specification](https://tools.ietf.org/html/rfc3986#page-17). A "Redirect URI is not valid" error will raise if this format is not respected.
 
-
 Go to the [Auth0 Dashboard](https://manage.auth0.com/#/applications), select your application and make sure that **Allowed Callback URLs** contains the following:
 
 #### iOS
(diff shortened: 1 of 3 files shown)


rafaelmaeuer:
Hello is anyone responsible for pull-requests? Do I need to open a new ticket for this? Can I get feedback from a maintainer please?

cocojoe:
Hi @rafaelmaeuer thanks for your contribution. If you look at the CI it gives you some errors, you should take a look at. https://jestjs.io/docs/en/snapshot-testing

lbalmaceda:
@rafaelmaeuer Thanks for your interest. Note this flow is not to be used on Public clients / mobile applications because it needs you to keep somewhere the application `client_secret`, which when talking mobile there's no safe place to store it 👍. What are you trying to accomplish? If calling your API is the answer please check this detailed guidance https://auth0.com/docs/architecture-scenarios/mobile-api.

rafaelmaeuer:
@cocojoe I will have a look at it, thanks for the advise.
@lbalmaceda yes my goal is to secure an API between a react native app and a server which doesn't make use of user profiles. So I thought the client credential grant is the optimal way to achieve this. As the secret could be stored encrypted in a secure space using [react-native-keychain](https://github.com/oblador/react-native-keychain) the app could be considered trusted or am I wrong? But you are right in one point, I didn't figured out yet, how to do the initial storing of that secret in the secure space without having it somewhere stored in the app. What grant would you recommend in my situation?

lbalmaceda:
@rafaelmaeuer Exactly. The link I shared has all the details and explanation for that scenario. 
> For authorizing a mobile app user and granting access to the API, Auth0 supports the Authorization Code Grant Flow with PKCE (see Proof Key for Code Exchange)

rafaelmaeuer:
So as I understand this correctly I need to use the [PKCE)](https://auth0.com/docs/architecture-scenarios/mobile-api/part-1#proof-key-for-code-exchange-pkce-) flow. Looking at the available methods in this package the only option I see, is to use `auth0.authorizeUrl` first to get an authorization_code and then use `auth0.exchange` to get an Access Token in exchange to the authorization_code. Is that correct?

rafaelmaeuer:
I need to make clear one more time that this app is NOT making use of user profiles. I don't want to use anything where a user needs to sing in or log in. I want general secure connection when doing API calls. Therefore I guess `webauth` is not suitable for me, as `* @param {String} [parameters.connection] The name of the identity provider to use, e.g. "google-oauth2" or "facebook". When not set, it will display Auth0's Universal Login Page.` will be required.

rafaelmaeuer:
Okay, first thanks for your all your advice! 

I found an Oauth seamless flow, proposed in this [ietf draft](https://tools.ietf.org/id/draft-hevroni-oauth-seamless-flow-00.html). Details about how the flow works is described in this [blogpost](https://blog.solutotlv.com/userless-mobile-authentication/). Would this be a solution to consider? 

I also thought about a solution where the device is authenticated instead of the user (e.g. using UUID and some generated password for login credentials. But this lacks of the verification process provided by e.g., double-opt-in using email.

lbalmaceda:
@rafaelmaeuer I've spoken with my colleagues and besides recommending what Auth0 currently offers and supports as part of the authorization product for this specific scenario I can't mention any other alternatives. Going with a custom solution to handle anonymous logins might be risky and should be avoided. What and how you implement it is on you and we cannot guarantee it's going to be safe. You should stick to the grant I mentioned earlier.

Regarding that seamless grant, I was told it didn't receive enough attention or in-deep reviews in terms of security and I wouldn't recommend using it either. If you have any further inquiries I recommend you reach our awesome team at https://support.auth0.com.

PR #213, opened 2019-04-16
Update telemetry format

lbalmaceda:
### Changes

- Changed telemetry format: use `env` to include the library version as the `react-native-auth0` key.
- Updated dev dependencies for fetch-mock to v7. Fixed snapshots.

### Testing

* [x] This change has been tested on the latest version of the platform/language or why not

### Checklist


auth/__tests__/__snapshots__/index.spec.js.snap
```diff
@@ -12,7 +12,7 @@ exports[`auth change password should return successful response 1`] = `Object {}
 
 exports[`auth change password should send correct payload 1`] = `
 Array [
-  "https://samples.auth0.com/dbconnections/change_password",
+  "https://samples.auth0.com/dbconnections/change_password?auth0Client=eyJuYW1lIjoicmVhY3QtbmF0aXZlLWF1dGgwIiwidmVyc2lvbiI6IjEuMC4wIn0%3D",
```
joshcanhelp: This should be a header, not a URL parameter. URL parameters are for redirects.

management/__tests__/__snapshots__/users.spec.js.snap
```diff
@@ -119,7 +119,7 @@ Object {
 
 exports[`users GET user should send correct payload 1`] = `
 Array [
-  "https://samples.auth0.com/api/v2/users/auth0%7C53b995f8bce68d9fc900099c",
+  "https://samples.auth0.com/api/v2/users/auth0%7C53b995f8bce68d9fc900099c?auth0Client=eyJuYW1lIjoicmVhY3QtbmF0aXZlLWF1dGgwIiwidmVyc2lvbiI6IjEuMC4wIn0%3D",
```
joshcanhelp: This should be a header, not a URL parameter. URL parameters are for redirects.

networking/__tests__/telemetry.spec.js
```diff
@@ -1,6 +1,6 @@
 import defaults from '../telemetry';
 
-describe('telemetry', () => {
+describe('basic telemetry', () => {
   it('should have name', () => {
```
joshcanhelp: Shouldn't it check for a version here as well?

lbalmaceda: You mean the test case that is below this one? This telemetry file defines name and version for this sdk, and does not include anything else. This is used from the telemetry value generator to obtain these details.

PR #223, opened 2019-07-02
Enable WebAuth Logout for Android & Fix iOS Logout.

lbalmaceda:
### Changes

This PR enables the clear session feature for Android and fixes it for iOS.

When this is merged, it will **always** send the `client_id` and the `returnTo` parameters when opening the Auth0 logout URL. 

Devs would need to whitelist the `returnTo` URL in the "Allowed Logout URLs" section of their **application's settings** in the Auth0 dashboard. Fail to do so will show an error page when the redirect is attempted.

#### Sample usage:

```js
auth0.webAuth
   .clearSession({})
   .then(success => {
       //Logged out!
    })
    .catch(error => {
       //Browser closed by the user.
     });
```

### Breaking changes?
At first, seems to only affect _iOS only_ (because Android was disabled!). But if we look deeper, the current behavior is to open the logout URL without passing any `returnTo` nor `clientId` values. Testing this in an iOS simulator resulted in a browser showing the text "OK" and not redirecting back to the app. So even if the log out was successful, I had to manually close the browser in order to get back to the app.

With the changes introduced in this PR, users will either face the "URL not whitelisted" screen when there's a dashboard configuration error or return successfully to the app after logging out.

#### Current iOS behavior:

http://recordit.co/TQqX1AyApp

#### iOS behavior after this PR:

http://recordit.co/gWa1dlR8Zp

#### Android behavior after this PR:
http://recordit.co/GcPBbvQeNh

### Testing

* [x] This change has been tested on the latest version of the platform/language or why not

### Checklist

### DIFF
--- a/webauth/__mocks__/auth0.js
+++ b/webauth/__mocks__/auth0.js
@@ -24,4 +24,6 @@ export default class A0Auth0 {
   oauthParameters(callback) {
     callback(this.parameters);
   }
+
+  bundleIdentifier = 'com.My.App';
 }
--- a/webauth/__mocks__/react-native.js
+++ b/webauth/__mocks__/react-native.js
@@ -5,5 +5,6 @@ const mock = {};
 
 mock.Linking = new Linking();
 mock.NativeModules = { A0Auth0: new A0Auth0() };
+mock.Platform = { OS: 'test-os' };
 
 module.exports = mock;
--- a/webauth/__tests__/webauth.spec.js
+++ b/webauth/__tests__/webauth.spec.js
@@ -1,0 +1,50 @@
+jest.mock('react-native');
+import Auth from '../../auth';
+import WebAuth from '../index';
+import { NativeModules } from 'react-native';
+import { URL } from 'url';
+
+const A0Auth0 = NativeModules.A0Auth0;
+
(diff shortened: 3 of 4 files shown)


lbalmaceda:
@stevehobbsdev I think that's how iOS behaves when trying to open a link from the app. @cocojoe can you clarify that, please?

damieng:
Force-merging as codecov is having issues again.

bneigher:
@lbalmaceda 
Guys, the "..wants to use Auth0 to Sign In" when using `clearSession` is preventing me from using the method. It's really weird to see that when a user chooses to logout.
Is there something I am missing that needs to be set up to skip that dialog?

cocojoe:
@bneigher Although I totally agree it isn't the most beautiful experience. This behaviour is part of Apple's functionality for the recommended way to implement Auth flow, the only way to clear the cookies is by using the same flow so the correct cookie jar is used. So the dialog will be presented and there is no way to disable it.

However, be great for you to feedback on this at https://auth0.com/feedback

nwaughachukwuma:
@cocojoe @lbalmaceda could the logout dialog message read something else? Say `You are going to be logged out from {app name}` or something

Issue #226, opened 2019-07-11
'App' wants to use 'app.com' to Sign In


yvocilon:
![Screenshot 2019-07-11 at 16 19 51](https://user-images.githubusercontent.com/38035825/61058702-09043b80-a3f8-11e9-9adc-f9c03ec0b3d7.png)

I've been getting the above message since upgrading to 1.4.2. I'm getting the messaging when using `auth0.webAuth.authorize` and `auth0.webAuth.clearSession`. 

Is it possible to disable this message or is there an explanation. I've been searching for quite a while but have not been able to find it. Or is this iOS specific and does it have nothing to do with this package? Our current app is running on a rather old version and the message does not show up there. 

Thank you!

lbalmaceda:
I found that as frustrating as you but it's not because of this package. That's iOS specific and cannot be customized nor disabled, AFAIK. Is something introduced in iOS10 and iOS11, but not present on iOS9 (that's why you don't see it on older devices)

yvocilon:
Ah okay. I upgraded our react-native-auth0 from 1.2.2 to 1.4.2 and did not have this issue with 1.2.2. Tested it on the same device.

wincod75:
@lbalmaceda Do you know if it's possible to change the "App Name" key that this alert points to? I have rebranded my app, changing the display name, but keeping the identifier. So it might be confusing if "AppName 1" wants to use Auth0, when everything is branded "AppName 2"

#258

PR #227, opened 2019-07-15
Disable authenticated session

pettersamuelsen:
### Changes
- Added an optional `useLegacyAuthentication` to the `authorize()` function
- `useLegacyAuthentication` will be `false` by default
- Fixes https://github.com/auth0/react-native-auth0/issues/226 when `ASWebAuthenticationSession` or `SFAuthenticationSession` is not needed 

### References

- https://github.com/auth0/react-native-auth0/issues/226

### Testing

* [x] This change has been tested on the latest version of the platform/language

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.java
@@ -53,7 +53,7 @@ public class A0Auth0Module extends ReactContextBaseJavaModule implements Lifecyc
     }
 
     @ReactMethod
-    public void showUrl(String url, boolean closeOnLoad, Callback callback) {
+    public void showUrl(String url, boolean closeOnLoad, boolean disableAuthenticatedSession Callback callback) {
         final Activity activity = getCurrentActivity();
 
         this.callback = callback;
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.java
@@ -53,7 +53,7 @@ public class A0Auth0Module extends ReactContextBaseJavaModule implements Lifecyc
     }
 
     @ReactMethod
-    public void showUrl(String url, boolean closeOnLoad, boolean disableAuthenticatedSession Callback callback) {
+    public void showUrl(String url, boolean closeOnLoad, boolean useLegacyAuthentication, Callback callback) {
         final Activity activity = getCurrentActivity();
 
         this.callback = callback;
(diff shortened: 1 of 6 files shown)


cocojoe:
@pettersamuelsen Please take a look at this for naming https://github.com/auth0/Auth0.swift/blob/master/Auth0/WebAuth.swift#L185

Also should be documented and called out as the real use case here is for non SSO.

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -53,7 +53,7 @@ public String getName() {
     }
 
     @ReactMethod
-    public void showUrl(String url, boolean closeOnLoad, Callback callback) {
+    public void showUrl(String url, boolean closeOnLoad, boolean disableAuthenticatedSession Callback callback) {
```
lbalmaceda: Isn't this missing a `,`?

webauth/agent.js
```diff
@@ -17,16 +17,21 @@ export default class Agent {
...
+      NativeModules.A0Auth0.showUrl(
+        url,
+        closeOnLoad,
+        disableAuthenticatedSession,
+        (error, redirectURL) => {
+          Linking.removeEventListener('url', urlHandler);
+          if (error) {
+            reject(error);
+          } else if (redirectURL) {
+            resolve(redirectURL);
```
lbalmaceda: what about joining the last 2 cases like this?

```
else if (redirectURL || closeOnLoad) {
  resolve(redirectURL);
}
```

webauth/index.js
```diff
@@ -52,7 +52,7 @@ export default class WebAuth {
    *
    * @memberof WebAuth
    */
-  authorize(options = {}) {
+  authorize({ disableAuthenticatedSession, ...options } = {}) {
```
lbalmaceda: Why do you need to define it here? Assume is present inside the `options` object and that's it. This is only used on iOS and this method is for a generic webauth.

android/src/main/java/com/auth0/react/A0Auth0Module.java
lbalmaceda: comma missing between params.

webauth/agent.js
pettersamuelsen: This code is untouched in this PR, so I'll leave it outside of the scope of this PR

pettersamuelsen:
@lbalmaceda  @cocojoe  I agree with your comments - see the updated commits

cocojoe:
Have had a think about this PR and I don't think there is a real need to support this. The flow employed is correct and on the iOS side is part of Apple's guidance and is the path going forward.

Issue #251, opened 2019-10-28
[2.1.0] TypeError: undefined is not a function, js engine: hermes


Bardiamist:
I have error `TypeError: undefined is not a function, js engine: hermes` after update from `2.0.0` to `2.1.0`. Only on Android and only if Debug is off.

Update: release apk also crash on start.

Minimum steps to reproduce:

1. `react-native init` (I tested on version 0.61.2 and 0.61.3 version)
2. Enable Hermes (one line change, see documentation)
3. `import 'react-native-auth0';`

joshcanhelp:
@Bardiamist - Thanks for the report here and apologies for the trouble. Can you give us a bit more to go on here? Possibly sample code that's causing the issue or a line number where that error is coming from?

Bardiamist:
Just error on start.
![](https://i.imgur.com/TgAT0cp.png)

But I can see what error in `jssrssign/lib/jssrssign.js` (thanks Reactotron)
![](https://i.imgur.com/ZBveZL5.png)

export-mike:
@Bardiamist I've downgraded to 2.0.0 for hermes support.

GuillaumeAmat:
For the record:

* `react-native-auth0@2.1.0` + Hermes = ❌
* `react-native-auth0@2.1.0` + Hermes + Debug mode = 👌

export-mike:
hmmm @2.1.0 is not going to work either.

jsrsasign is a dependency

2.1.0
```
npm list jsrasign 
   ┬ react-native-auth0@2.1.0
  └── jsrsasign@8.0.12
```

2.0.0
```
 npm list jsrsasign
/Users/mike/projects/temp-app
└── (empty)
```

cocojoe:
@Widcket can you take a look at this please

Widcket:
Thanks everyone for the information. This issue is in our radar, and we’re currently evaluating options.

ix42:
Having the same problem with 2.1.0. Only downgrading to 2.0.0 fixed it

lbalmaceda:
@Bardiamist & all, I've put a PR that replaces jsrsasign. I've tested it and an Hermes enabled app works fine now. We'll be making a new release next week but in the meantime I'd love you have your feedback. 

You can try the branch code by installing the package this way:


```
yarn add https://github.com/auth0/react-native-auth0.git#temp-idtv-2
```

Bardiamist:
I checked, that works for my application.

[...]

PR #268, opened 2020-01-16
Refactor RSA verification: Replace jsrsasign with crypto-js

lbalmaceda:
### Changes

`jsrsasign` library was not compatible with the Hermes JS engine, currently optional but still officially supported by the React Native platform for Android. This PR changes the RSA signature check implementation to use crypto-js instead.

No breaking changes are introduced on this build.

### References
Resolves https://github.com/auth0/react-native-auth0/issues/251

### Testing

The unit tests were updated. To evaluate Hermes support, it has been tested this way:
1. Enable Hermes for your android app by changing the `app/build.gradle` file 

```groovy
 project.ext.react = [
     entryFile: "index.js",
     enableHermes: true,  // clean and rebuild if changing
 ]
```
2. Clean and Rebuild
```sh
cd android && ./gradlew clean assemble
```

3. Run the android app in a "release" variant
```sh
react-native run-android --variant release
```

### Checklist


package.json
```diff
@@ -48,7 +48,8 @@
   },
   "dependencies": {
     "base-64": "^0.1.0",
-    "jsrsasign": "8.0.12",
+    "crypto-js": "^3.1.9-1",
+    "jsbn": "^1.1.0",
     "jwt-decode": "^2.2.0",
```
lbalmaceda: improvement: consider removing `jwt-decode`, add a base64-url-decode method to the base64 class introduced here and do it ourselves

src/jwt/__tests__/jwt.spec.js
```diff
@@ -88,30 +89,15 @@ describe('id token verification tests', () => {
...
-      setupFetchMock({jwks});
+      setupFetchMock();
 
       await expect(verify(testJwt)).resolves.toBeUndefined();
     });
 
     const setupFetchMock = ({
       domain = BASE_EXPECTATIONS.domain,
-      jwks = getJwks(),
+      jwks = getExpectedJwks(),
```
lbalmaceda: renamed to reflect this is the "happy path"

src/jwt/__tests__/jwt.spec.js
```diff
@@ -114,30 +115,15 @@ describe('id token verification tests', () => {
...
-      };
-
-      setupFetchMock({jwks});
+      setupFetchMock();
 
       await expect(verify(testJwt)).resolves.toBeUndefined();
     });
 
     const setupFetchMock = ({
       domain = BASE_EXPECTATIONS.domain,
```
Widcket: Despite hoisting, shouldn't this be put at the top, or at the bottom (or in a helpers file), instead of being in the middle of the file?

src/jwt/base64.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Borrowed from IDToken-verifier package
+ * https://github.com/auth0/idtoken-verifier/blob/master/src/helpers/base64.js
+ */
+import base64 from 'base64-js';
+
+export function padding(str) {
+  var mod = str.length % 4;
```
Widcket: Shouldn't that `4` be in a constant with a descriptive name?

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+
+import {BigInteger} from 'jsbn';
+import SHA256 from 'crypto-js/sha256';
+
+var DigestInfoHead = {
+  sha1: '3021300906052b0e03021a05000414',
+  sha224: '302d300d06096086480165030402040500041c',
+  sha256: '3031300d060960864801650304020105000420',
+  sha384: '3041300d060960864801650304020205000430',
+  sha512: '3051300d060960864801650304020305000440',
```
Widcket: Why include entries for algorithms we do not support?

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+
+var DigestAlgs = {
+  sha256: SHA256,
+};
+
+function RSAVerifier(modulus, exp) {
+  this.n = null;
+  this.e = 0;
+
+  if (modulus != null && exp != null && modulus.length > 0 && exp.length > 0) {
```
Widcket: Shouldn't this be checking for truthy values instead of `!= null`? Like `if (modulus && exp && ...)`

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+    this.n = new BigInteger(modulus, 16);
+    this.e = parseInt(exp, 16);
+  } else {
+    throw new Error('Invalid key data');
```
Widcket: Is this error message used somewhere?

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
+/*
+Based on the work of Tom Wu
+http://www-cs-students.stanford.edu/~tjw/jsbn/
+http://www-cs-students.stanford.edu/~tjw/jsbn/LICENSE
+*/
+
+import {BigInteger} from 'jsbn';
+import SHA256 from 'crypto-js/sha256';
+
+var DigestInfoHead = {
```
Widcket: Shouldn't this var use camelCase?

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+  sha256: '3031300d060960864801650304020105000420',
+  sha384: '3041300d060960864801650304020205000430',
+  sha512: '3051300d060960864801650304020305000440',
+  md2: '3020300c06082a864886f70d020205000410',
+  md5: '3020300c06082a864886f70d020505000410',
+  ripemd160: '3021300906052b2403020105000414',
+};
+
+var DigestAlgs = {
+  sha256: SHA256,
```
Widcket: Shouldn't this var use camelCase?

src/jwt/rsa-verifier.js
Widcket: And why `var`? Isn't `const` more appropriate since it doesn't get reassigned?

src/jwt/rsa-verifier.js
Widcket: And why `var`? Isn't `const` more appropriate since it doesn't get reassigned?

src/jwt/base64.js
```diff
@@ -0,0 +1,31 @@
...
+  var pad = 4 - mod;
+
```
Widcket: `mod` and `pad` should be `const`, not `var`.

src/jwt/base64.js
```diff
@@ -0,0 +1,31 @@
...
+    return str;
+  }
+
+  return str + new Array(1 + pad).join('=');
+}
+
+function byteArrayToHex(raw) {
+  var HEX = '';
+
+  for (var i = 0; i < raw.length; i++) {
```
Widcket: `let` could be used for `i` instead of `var`.

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+  if (modulus != null && exp != null && modulus.length > 0 && exp.length > 0) {
+    this.n = new BigInteger(modulus, 16);
+    this.e = parseInt(exp, 16);
+  } else {
+    throw new Error('Invalid key data');
+  }
+}
+
+function getAlgorithmFromDigest(hDigestInfo) {
+  for (var algName in DigestInfoHead) {
```
Widcket: `algName` could be `const` and not `var`.

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+    var head = DigestInfoHead[algName];
+    var len = head.length;
```
Widcket: `len` and `head` should be `const` and not `var`.

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+        alg: algName,
+        hash: hDigestInfo.substring(len),
+      };
+    }
+  }
+  return [];
+}
+
+RSAVerifier.prototype.verify = function(msg, encsig) {
+  encsig = encsig.replace(/[^0-9a-f]|[\s\n]]/gi, '');
```
Widcket: Probably not a good idea to mutate a parameter.

Widcket: Shouldn't the parameter be in camelCase?

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+
+  var sig = new BigInteger(encsig, 16);
+  if (sig.bitLength() > this.n.bitLength()) {
+    throw new Error('Signature does not match with the key modulus.');
+  }
+
+  var decryptedSig = sig.modPowInt(this.e, this.n);
+  var digest = decryptedSig.toString(16).replace(/^1f+00/, '');
+
+  var digestInfo = getAlgorithmFromDigest(digest);
```
Widcket: `sig`, `decryptedSig`, `digest` and `digestInfo` should be `const` instead of `var`.

src/jwt/rsa-verifier.js
```diff
@@ -0,0 +1,76 @@
...
+  if (digestInfo.length === 0) {
+    return false;
+  }
+
+  if (!DigestAlgs.hasOwnProperty(digestInfo.alg)) {
+    throw new Error('Hashing algorithm is not supported.');
+  }
+
+  var msgHash = DigestAlgs[digestInfo.alg](msg).toString();
```
Widcket: `msgHash` should be a `const`.

src/jwt/signatureVerifier.js
```diff
@@ -44,21 +45,30 @@ export const verifySignature = (idToken, options) => {
...
 
-  return getJwk(options.domain, header.kid).then(jwk => {
-    const pubKey = KEYUTIL.getKey(jwk);
-    const signatureValid = KJUR.jws.JWS.verify(idToken, pubKey, ['RS256']);
+  return getJwk(options.domain, header.kid)
+    .then(jwk => rsaVerifierForKey(jwk))
+    .then(rsaVerifier => {
+      const encodedParts = idToken.split('.');
+      const headerAndPayload = encodedParts[0] + '.' + encodedParts[1];
+      const signature = base64.decodeToHEX(encodedParts[2]);
```
Widcket: Is there some prior verification that the ID Token has 3 parts?

src/jwt/signatureVerifier.js
```diff
@@ -44,21 +45,30 @@ export const verifySignature = (idToken, options) => {
...
+          desc: 'Invalid ID token signature',
+        }),
+      );
+    });
+};
+
+const rsaVerifierForKey = jwk => {
+  const modulus = base64.decodeToHEX(jwk.n);
+  const exponent = base64.decodeToHEX(jwk.e);
+  return Promise.resolve(new RSAVerifier(modulus, exponent));
```
Widcket: Why make a Promise for this non-async operation?

src/jwt/__tests__/jwt.spec.js
```diff
@@ -12,16 +11,6 @@ describe('id token verification tests', () => {
       fetchMock.restore();
     });
 
-    it('uses fixed version of jsrsasign', () => {
```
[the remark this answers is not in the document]
lbalmaceda: Good point. I'll add that

src/jwt/base64.js
lbalmaceda: I've taken these files (and their tests) from `auth0.js` as-is, as the header on this file points out. I don't think I should be moddifying them.

package.json
```diff
@@ -48,7 +48,8 @@
   },
   "dependencies": {
     "base-64": "^0.1.0",
-    "jsrsasign": "8.0.12",
+    "crypto-js": "^3.1.9-1",
```
lbalmaceda: note to self, pin this version

lbalmaceda: No need to pin. Keeping as-is will enable receiving patch releases 👍

lbalmaceda: Won't do. Will keep patch releases 👍

src/jwt/__tests__/jwt.spec.js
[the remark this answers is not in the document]
lbalmaceda: I decided to keep accepting patch versions. This test will not be added back.

src/jwt/__tests__/jwt.spec.js
lbalmaceda: I'll see if I can move it upwards 👍 thanks

src/jwt/signatureVerifier.js
lbalmaceda: Yes, it's done at the beginning of `verifySignature` when jwtDecoder is called

src/jwt/signatureVerifier.js
lbalmaceda: Will refactor ⚡️

src/jwt/__tests__/jwt.spec.js
lbalmaceda: Moved to the bottom

src/jwt/rsa-verifier.js
lbalmaceda: I'll remove the ones we don't support. But I think it's worth keeping the supporting structure around (the 2 arrays)

src/jwt/rsa-verifier.js
lbalmaceda: I'm not 100% sure. We will need to check.

src/jwt/rsa-verifier.js
lbalmaceda: Do you mean when this is thrown? If so, is when the key arguments are invalid and lead to an impossible key instantiation.

src/jwt/signatureVerifier.js
lbalmaceda: I think this was like this to be able to reject the promise if there was a "runtime error" on the verifier instantiation. We will need to check.

src/jwt/rsa-verifier.js
lbalmaceda: This _message_ is not used. However, the error needs to be thrown in order to catch it later (on this same promise chain) and reject the Promise with the same "key not found" exception when the key id is missing or the request failed. Avoids duplicating the custom exception instantiation and Promise rejection.

src/jwt/rsa-verifier.js
lbalmaceda: Correct. Changed.

src/jwt/__tests__/base64.spec.js
```diff
@@ -0,0 +1,17 @@
...
+    expect(base64.padding('')).toBe('');
+    expect(base64.padding('a')).toBe('a===');
+    expect(base64.padding('ab')).toBe('ab==');
+    expect(base64.padding('abc')).toBe('abc=');
+    expect(base64.padding('abcd')).toBe('abcd');
+    expect(base64.padding('abced')).toBe('abced===');
+    expect(base64.padding(base64.padding('abc'))).toBe('abc=');
+  });
+
+  it('decode to hex', function() {
```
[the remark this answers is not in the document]
lbalmaceda: I don't see other choices here. The method's responsibility is pretty dumb, TBH. Only decoding a base64 into a HEX value. What do you have in mind?

PR #269, opened 2020-01-16
Handle missing kid (key id) on the JWKS

lbalmaceda:
### Changes

Added a missing use case in which the public key the SDK requires can't be found

### Testing

### Checklist


src/jwt/signatureVerifier.js
```diff
@@ -71,13 +71,16 @@ const getJwk = (domain, kid) => {
           k => k.use === 'sig' && k.kty === 'RSA' && k.kid && (k.n && k.e),
         )
         .find(k => k.kid === kid);
+      if (!key) {
+        throw new Error('Key not present');
```
lbalmaceda: Doesn't really matter what message we put here, as it will get caught on the "catch" clause below

PR #270, opened 2020-01-22
Add passwordless endpoints

lbalmaceda:
### Changes

This PR adds support for invoking the new passwordless endpoints. It also adds a small note on the readme about them and the grant type that must be enabled first.

There are some UX changes on the second request (the one that performs the authentication). The endpoint expects a `username` parameter, but I'm mapping it from either `email` or `phoneNumber`. The same happens with `otp`, which is getting mapped from `code`. Please, evaluate if this is desired or if it's only adding noise.

> Any formatting change included on the diff is due to the pre-commit hook.

### References

`SDK-1281`

### Testing

I've tested the email passwordless manually on a sample app. Unit tests are included as well.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,58 @@ auth0.auth
   .catch(console.error);
 ```
 
+#### Login with Passwordless
+
+Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application before-hand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
+
+To start the flow, you request a link or code to be sent to the user's email or phone number:
+
+```js
+auth0.auth
+  .passwordlessWithEmail({
+    email: 'info@auth0.com',
+    send: 'link',
+  })
+  .then(console.log)
+  .catch(console.error);
+```
+
+or
+
+```js
+auth0.auth
+  .passwordlessWithSMS({
+    phoneNumber: '+5491159991000',
+    send: 'code',
+  })
+  .then(console.log)
+  .catch(console.error);
+```
+
+Then, in order to complete the authentication, you must send back that received code value along with the email or phone number used:
+
+```js
+auth0.auth
+  .loginWithEmail({
+    email: 'info@auth0.com',
+    code: '123456',
+  })
+  .then(console.log)
+  .catch(console.error);
+```
+
+or
+
+```js
+auth0.auth
+  .loginWithSMS({
+    phoneNumber: 'info@auth0.com',
+    code: '123456',
+  })
+  .then(console.log)
+  .catch(console.error);
+```
+
 #### Create user in database connection
 
 ```js
--- a/README.md
+++ b/README.md
@@ -256,9 +256,9 @@ auth0.auth
 
 #### Login with Passwordless
 
-Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application beforehand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
+Passwordless is a two-step authentication flow that makes use of this type of connection. The **Passwordless OTP** grant is required to be enabled in your Auth0 application beforehand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
 
-To start the flow, you request a link or code to be sent to the user's email or phone number:
+To start the flow, you request a code to be sent to the user's email or phone number. For email scenarios only, a link can be sent in place of the code.
 
 ```js
 auth0.auth
@@ -276,7 +276,6 @@ or
 auth0.auth
   .passwordlessWithSMS({
     phoneNumber: '+5491159991000',
-    send: 'code',
   })
   .then(console.log)
   .catch(console.error);
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ auth0.auth
 
 #### Login with Passwordless
 
-Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application before-hand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
+Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application beforehand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
 
 To start the flow, you request a link or code to be sent to the user's email or phone number:
 
--- a/src/auth/__tests__/index.spec.js
+++ b/src/auth/__tests__/index.spec.js
@@ -160,7 +160,7 @@ describe('auth', () => {
     });
   });
 
-  describe.only('passwordless flow', () => {
+  describe('passwordless flow', () => {
     describe('with email connection', () => {
       it('should begin with code and authParams', async () => {
         fetchMock.postOnce(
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -254,6 +254,58 @@ auth0.auth
   .catch(console.error);
 ```
 
+#### Login with Passwordless
+
+Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application before-hand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
```
Widcket:
```suggestion
Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application beforehand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
```

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
-        scope: 'openid'
+        scope: 'openid',
       };
       await expect(auth.exchange(parameters)).rejects.toMatchSnapshot();
     });
   });
 
+  describe('passwordless flow', () => {
+    describe('with email connection', () => {
+      it('should begin with code and authParams', async () => {
```
Widcket: You might want to have a 'begin' test just with the code and test the `authParams` separately.

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
+        fetchMock.postOnce(
+          'https://samples.auth0.com/passwordless/start',
```
Widcket: Since this url is being used twice, you might want to put it in a `const` below the nearest `describe`, or wrap that mock in a function call.

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
+        expect.assertions(1);
+        await auth.passwordlessWithEmail({
+          email: 'info@auth0.com',
+          send: 'link',
+        });
+        expect(fetchMock.lastCall()).toMatchSnapshot();
+      });
+
+      it('should continue', async () => {
+        fetchMock.postOnce('https://samples.auth0.com/oauth/token', tokens);
```
Widcket: Idem, see comment above.

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
         state,
-        scope: 'openid'
+        scope: 'openid',
       };
       await expect(auth.exchange(parameters)).rejects.toMatchSnapshot();
     });
   });
 
+  describe('passwordless flow', () => {
+    describe('with email connection', () => {
```
Widcket: Shouldn't this be `context`?

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
+          code: '123456',
+          audience: 'http://myapi.com',
+          scope: 'openid',
+        });
+        expect(fetchMock.lastCall()).toMatchSnapshot();
+      });
+    });
+
+    describe('with SMS connection', () => {
+      it('should begin with code and authParams', async () => {
```
Widcket: You might want to have a 'begin' test just with the code and test the authParams separately.

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
+          email: 'info@auth0.com',
+          code: '123456',
+          audience: 'http://myapi.com',
+          scope: 'openid',
+        });
+        expect(fetchMock.lastCall()).toMatchSnapshot();
+      });
+    });
+
+    describe('with SMS connection', () => {
```
Widcket: Shouldn't this be `context`?

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,143 @@ describe('auth', () => {
...
+          authParams: {
+            scope: 'openid profile',
+          },
+        });
+        expect(fetchMock.lastCall()).toMatchSnapshot();
+      });
+
+      it('should begin with link', async () => {
+        fetchMock.postOnce(
+          'https://samples.auth0.com/passwordless/start',
```
Widcket: Idem as https://github.com/auth0/react-native-auth0/pull/270/files#r370146316, and all the instances below.

src/auth/index.js
```diff
@@ -169,17 +169,147 @@ export default class Auth {
...
+      })
+      .then(responseHandler);
+  }
+
+  /**
+   * Starts the Passworldess flow with an SMS connection
+   *
+   * @param {Object} parameters passwordless parameters
+   * @param {String} parameters.phoneNumber the phone number to send the link/code to
+   * @param {String} parameters.send the passwordless strategy, either 'link' or 'code'
```
Widcket: Isn't `link` not allowed when using SMS? See https://auth0.com/docs/api/authentication#get-code-or-link:

> You have three options for passwordless authentication:
> 
> - Send a verification code using email.
> - Send a link using email.
> - Send a verification code using SMS.

src/auth/index.js
```diff
@@ -169,17 +169,147 @@ export default class Auth {
...
+   * @param {String} parameters.authParams optional parameters, used when strategy is 'linḱ'
```
Widcket: Idem

src/auth/index.js
```diff
@@ -169,17 +169,147 @@ export default class Auth {
...
+        grant_type: 'http://auth0.com/oauth/grant-type/passwordless/otp',
+      })
+      .then(responseHandler);
+  }
+
+  /**
+   * Finishes the Passworldess authentication with an SMS connection
+   *
+   * @param {Object} parameters passwordless parameters
+   * @param {String} parameters.phoneNumber the phone number where the link/code was received
```
Widcket: Same as above.

src/auth/index.js
```diff
@@ -169,17 +169,147 @@ export default class Auth {
...
+        },
+      },
+      parameters,
+    );
+    return this.client
+      .post('/oauth/token', {
+        ...payload,
+        client_id: this.clientId,
+        realm: 'sms',
+        grant_type: 'http://auth0.com/oauth/grant-type/passwordless/otp',
```
Widcket: You might want to extract this, to avoid repeating it in https://github.com/auth0/react-native-auth0/pull/270/files?file-filters%5B%5D=.js&file-filters%5B%5D=.md#diff-2b0a2893c010a4aea8147a48a8fdd092R278

src/auth/index.js
```diff
@@ -314,20 +444,20 @@ export default class Auth {
...
-        }
+          email: {required: true},
+          password: {required: true},
+          connection: {required: true},
+          username: {required: false},
+          metadata: {required: false, toName: 'user_metadata'},
+        },
       },
-      parameters
+      parameters,
```
Widcket: Do the style conventions for this repo allow trailing commas?

Widcket:
> There are some UX changes on the second request (the one that performs the authentication). The endpoint expects a username parameter, but I'm mapping it from either email or phoneNumber. The same happens with otp, which is getting mapped from code. Please, evaluate if this is desired or if it's only adding noise.

I think this is good from a UX perspective.

src/auth/__tests__/index.spec.js
lbalmaceda: I don't see it being used anywhere else. It could be true for mocha tests but I don't think it's worth refactoring every test file now.

src/auth/__tests__/index.spec.js
lbalmaceda: It's a good refactor but the rest of the tests on this file at least, are doing the same.

src/auth/index.js
lbalmaceda: That's a good question actually. I don't know the answer. But on [this doc](https://auth0.com/docs/connections/passwordless/guides/sms-otp#configure-passwordless-sms-settings) they still mention link as a possible strategy for SMS connections

> Only the last one-time password **(or link)** issued will be accepted.

Let me ask the right team and come back

src/auth/index.js
lbalmaceda: Yes, enforced via pre-commit hook https://github.com/auth0/react-native-auth0/blob/master/.prettierrc.js#L5

src/auth/index.js
lbalmaceda: This particular endpoint is called for many different flows (5 calls on this file currently). What mostly changes all the time is the grant_type sent. But there are also additional parameters, such as connection/realm, and the additional values the user could include. I'm afraid that moving it into a function would maintain the same structure, as we need to pass those arguments anyway. Worth to revisit later.

src/auth/index.js
lbalmaceda: Remove this param from the doc-block.

src/auth/__tests__/index.spec.js
```diff
@@ -133,35 +138,169 @@ describe('auth', () => {
...
+        );
+        expect.assertions(1);
+        await auth.passwordlessWithSMS({
+          phoneNumber: '+5491159991000',
+          send: 'code',
+        });
+        expect(fetchMock.lastCall()).toMatchSnapshot();
+      });
+
+      it('should begin with link', async () => {
```
Widcket: If we're removing the link from SMS, we should remove this test as well.

lbalmaceda: We're not removing support for `link`. We're just not documenting it 👍 Tests should still prove the endpoint can be called with additional auth params and/or with link

src/auth/index.js
Widcket: The way I've seen that Auth0.swift solves it, is to create a custom matcher (a function, in this case) for every use case instead of every endpoint. So it has many matchers for `/oauth/token`, each one configured differently for each use case.

Widcket: And each of these matchers may or may not take params, according to the needs of the use case.

README.md
```diff
@@ -254,6 +254,58 @@ auth0.auth
   .catch(console.error);
 ```
 
+#### Login with Passwordless
+
+Passwordless is a two-step authentication flow, and requires the **Passwordless OTP** grant to be enabled in your Auth0 application beforehand. Check [our guide](https://auth0.com/docs/dashboard/guides/applications/update-grant-types) to learn how to enable it.
+
+To start the flow, you request a link or code to be sent to the user's email or phone number:
```
Widcket: You might want to reword this to avoid implying that you can use a link with SMS.

src/auth/__tests__/index.spec.js
Widcket: But tests are documentation, and it may be confusing for a customer to come across this.

Issue #254, opened 2019-11-07
token exchange API used by Sign In With Apple is missing


linfan68:
### Describe the problem you'd like to have solved

I need to add Sign In With Apple in my React Native app

### Describe the ideal solution

adding API: tokenExchange(withAppleAuthorizationCode  similar to the SWIFT client

## Alternatives and current work-arounds

call the client post function directly:

```ts
const auth: any = (auth0.auth as any)
const payload = {
  client_id: auth.clientId,
  grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
  subject_token: credential.authorizationCode,
  subject_token_type: 'http://auth0.com/oauth/token-type/apple-authz-code',
  scope: 'openid profile offline_access',
}
const res = await auth.client.post('/oauth/token', payload)
```

lbalmaceda:
@cocojoe @Widcket please, evaluate

Widcket:
Hey @linfan68, thanks for raising this!

himalayaahuja:
> ### Describe the problem you'd like to have solved
> I need to add Sign In With Apple in my React Native app
> 
> ### Describe the ideal solution
> adding API: tokenExchange(withAppleAuthorizationCode similar to the SWIFT client
> 
> ## Alternatives and current work-arounds
> call the client post function directly:
> 
> ```ts
> const auth: any = (auth0.auth as any)
> const payload = {
>   client_id: auth.clientId,
>   grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
>   subject_token: credential.authorizationCode,
>   subject_token_type: 'http://auth0.com/oauth/token-type/apple-authz-code',
>   scope: 'openid profile offline_access',
> }
> const res = await auth.client.post('/oauth/token', payload)
> ```
> 
> ### Additional context

Is this correct and working ?

PR #273, opened 2020-02-05
Added token exchange for native social endpoint [SDK-1307]

Widcket:
### Changes

- Added the [Token Exchange for Native Social](https://auth0.com/docs/api/authentication#token-exchange-for-native-social) endpoint to the Auth API

### References

- Closes https://github.com/auth0/react-native-auth0/issues/254

### Testing

### Checklist

### DIFF
--- a/src/auth/__tests__/index.spec.js
+++ b/src/auth/__tests__/index.spec.js
@@ -160,6 +160,95 @@ describe('auth', () => {
     });
   });
 
+  describe('code exchange for native social', () => {
+    it('should send correct payload', async () => {
+      fetchMock.postOnce('https://samples.auth0.com/oauth/token', tokens);
+      expect.assertions(1);
+      await auth.exchangeNativeSocial({
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+      });
+      expect(fetchMock.lastCall()).toMatchSnapshot();
+    });
+
+    it('should send correct payload with optional parameters', async () => {
+      fetchMock.postOnce('https://samples.auth0.com/oauth/token', tokens);
+      expect.assertions(1);
+      await auth.exchangeNativeSocial({
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+        userProfile: {
+          name: {
+            firstName: 'John',
+            lastName: 'Smith',
+          },
+        },
+        audience: 'http://myapi.com',
+        scope: 'openid',
+      });
+      expect(fetchMock.lastCall()).toMatchSnapshot();
+    });
+
+    it('should return successful response', async () => {
+      fetchMock.postOnce('https://samples.auth0.com/oauth/token', tokens);
+      expect.assertions(1);
+      const parameters = {
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+      };
+      await expect(
+        auth.exchangeNativeSocial(parameters),
+      ).resolves.toMatchSnapshot();
+    });
+
+    it('should return successful response with optional parameters', async () => {
+      fetchMock.postOnce('https://samples.auth0.com/oauth/token', tokens);
+      expect.assertions(1);
+      const parameters = {
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+        userProfile: {
+          name: {
+            firstName: 'John',
+            lastName: 'Smith',
+          },
+        },
+        audience: 'http://myapi.com',
+        scope: 'openid',
+      };
+      await expect(
+        auth.exchangeNativeSocial(parameters),
+      ).resolves.toMatchSnapshot();
+    });
+
+    it('should handle oauth error', async () => {
+      fetchMock.postOnce('https://samples.auth0.com/oauth/token', oauthError);
+      expect.assertions(1);
+      const parameters = {
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+      };
+      await expect(
+        auth.exchangeNativeSocial(parameters),
+      ).rejects.toMatchSnapshot();
+    });
+
+    it('should handle unexpected error', async () => {
+      fetchMock.postOnce(
+        'https://samples.auth0.com/oauth/token',
+        unexpectedError,
+      );
+      expect.assertions(1);
+      const parameters = {
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+      };
+      await expect(
+        auth.exchangeNativeSocial(parameters),
+      ).rejects.toMatchSnapshot();
+    });
+  });
+
   describe('passwordless flow', () => {
     describe('with email connection', () => {
       it('should begin with code', async () => {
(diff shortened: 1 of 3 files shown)


src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   * Exchanges a code obtained via a native social authentication solution for the user's tokens
+   *
+   * @param {Object} parameters parameters used to obtain tokens from a code
+   * @param {String} parameters.subjectToken code returned by the native social authentication solution
+   * @param {String} parameters.subjectTokenType identifier that indicates the type of `subjectToken`
+   * @param {String} parameters.userProfile optional element used for native iOS interactions for which profile updates can occur. Only for `APPLE-AUTHZ-CODE`
+   * @param {String} parameters.audience optional API audience to request
+   * @param {String} parameters.scope optional scopes to request
+   * @returns {Promise}
+   *
```
lbalmaceda: add

```
@see https://auth0.com/docs/api/authentication#token-exchange-for-native-social
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
       .then(responseHandler);
   }
 
+  /**
+   * Exchanges a code obtained via a native social authentication solution for the user's tokens
```
lbalmaceda:
```suggestion
   * Exchanges an external token obtained via a native social authentication solution for the user's tokens
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   *
+   * @param {Object} parameters parameters used to obtain tokens from a code
```
lbalmaceda:
```suggestion
   * @param {Object} parameters parameters used to obtain user tokens from an external provider's token
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   * @param {String} parameters.subjectToken code returned by the native social authentication solution
```
lbalmaceda:
```suggestion
   * @param {String} parameters.subjectToken token returned by the native social authentication solution
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   * @param {String} parameters.subjectTokenType identifier that indicates the type of `subjectToken`
```
lbalmaceda:
```suggestion
   * @param {String} parameters.subjectTokenType identifier that indicates the native social authentication solution
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   * @param {String} parameters.userProfile optional element used for native iOS interactions for which profile updates can occur. Only for `APPLE-AUTHZ-CODE`
```
lbalmaceda: optionality is determined by the use of brackets when describing a parameter

```suggestion
   * @param {String} [parameters.userProfile] additional profile attributes to set or override, only on select native social authentication solutions
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   * @param {String} parameters.audience optional API audience to request
```
lbalmaceda:
```suggestion
   * @param {String} [parameters.audience] API audience to request
```

src/auth/index.js
```diff
@@ -118,6 +118,41 @@ export default class Auth {
...
+   * @param {String} parameters.scope optional scopes to request
```
lbalmaceda: I'd also display a sample value, just like the ones that exist today on other docs.

```suggestion
   * @param {String} [parameters.scope] scope to request, i.e. "openid profile email"
```

src/auth/__tests__/index.spec.js
```diff
@@ -160,6 +160,95 @@ describe('auth', () => {
...
+    });
+
+    it('should send correct payload with optional parameters', async () => {
+      fetchMock.postOnce('https://samples.auth0.com/oauth/token', tokens);
+      expect.assertions(1);
+      await auth.exchangeNativeSocial({
+        subjectToken: 'a subject token',
+        subjectTokenType: 'a subject token type',
+        userProfile: {
+          name: {
```
lbalmaceda: Is this the correct usage for SIWA, for example? Or is it all in a plain structure like

```
userProfile : {
  firstName: 'john',
  lastName: 'smith'
}
```

src/auth/index.js
Widcket: Why use an array here?

src/auth/index.js
Widcket: The return type is wrong also, will update it. It should't be an array, but an object.

src/auth/index.js
Widcket: Again, why use an array here?

src/auth/__tests__/index.spec.js
Widcket: For SIWA, according to the [API docs](https://auth0.com/docs/api/authentication#token-exchange-for-native-social), is:

> Expected parameter value will be JSON in the form of: { name: { firstName: 'John', lastName: 'Smith }}

src/auth/index.js
Widcket: Added in 0758eb8fbb8a59f43b77780b9bd125f46da6bf35. Thanks, great review BTW

src/auth/index.js
lbalmaceda: > optionality is determined by the use of brackets when describing a parameter

The square brackets are used in typescript definitions to mark a given parameter as optional. So instead of saying "optional parameter to..." you put the brackets and that's it.

Regarding types, those are explained by the curly braces. So for this line, it's saying that is `an optional String property named userProfile`.

Use the above to fix the remaining comments (the ones below this line)

lbalmaceda:
There are still comments to be addressed ⚡️

src/auth/index.js
Widcket: Fixed in b6cf3cb8110a349ce96ca14b5512c3e17a42fa57.

src/auth/index.js
Widcket: Fixed in b6cf3cb8110a349ce96ca14b5512c3e17a42fa57.

src/auth/index.js
Widcket: Fixed in b6cf3cb8110a349ce96ca14b5512c3e17a42fa57.

src/auth/index.js
```diff
@@ -118,6 +118,43 @@ export default class Auth {
...
 
+  /**
+   * Exchanges an external token obtained via a native social authentication solution for the user's tokens
+   *
+   * @param {Object} parameters parameters used to obtain user tokens from an external provider's token
+   * @param {String} parameters.subjectToken token returned by the native social authentication solution
+   * @param {String} parameters.subjectTokenType identifier that indicates the native social authentication solution
+   * @param {Object} [parameters.userProfile] additional profile attributes to set or override, only on select native social authentication solutions
+   * @param {String} [parameters.audience] API audience to request
+   * @param {String} [parameters.scope] scopes to request
```
lbalmaceda: As in other `scope` params from [this same file](https://github.com/auth0/react-native-auth0/blob/b6cf3cb8110a349ce96ca14b5512c3e17a42fa57/src/auth/index.js#L166), let's add an example:

```suggestion
   * @param {String} [parameters.scope] scopes requested for the issued tokens. e.g. `openid profile`
```

Issue #294, opened 2020-04-16
Getting signin popup during logout


lhuria94:
Hi, 

We are getting this popup during logout from the mobile app.
Our code is:
```
auth0.webAuth
      .clearSession({})
      .then(success => {
        Alert.alert('Logged out!');
        setAccessToken(null);
      })
      .catch(error => {
        console.log('Log out cancelled', error);
      });
```

Is there any way to avoid this situation?

Widcket:
Hi @lhuria94, can you share a screenshot of the popup?

mountainfirefly:
I am having the same issue, here is the screenshot @Widcket
![Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-17 at 15 15 46](https://user-images.githubusercontent.com/11595211/79556275-79926400-80be-11ea-8cbc-dae99f2934bf.png)

mountainfirefly:
![Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-17 at 15 15 46](https://user-images.githubusercontent.com/11595211/79556663-04735e80-80bf-11ea-8f38-f34dbb91686d.png)
I am having the same issue, here is the screenshot. @Widcket

Widcket:
@mountainfirefly thanks for the screenshot. Support for disabling SSO (and thus avoiding that popup) is already in our backlog.

lhuria94:
@Widcket But we need the SSO as well, even if we have the popup its the message which does not make sense for the. A signed in user has to sign in to logout.

Widcket:
@lhuria94 both popup and message are displayed by `ASWebAuthenticationSession`, and unfortunately we do not have control over it. However if you'd like to force the user to login again, you can add the `prompt: "login"` param to the login call.

pavelpashkovsky:
v2.3.1 still getting this popup. Is there any solution to avoid it except rolling back to old versions?

paulosborne:
@Widcket 

> @lhuria94 both popup and message are displayed by `ASWebAuthenticationSession`, and unfortunately we do not have control over it. However if you'd like to force the user to login again, you can add the `prompt: "login"` param to the login call.

Is there a way to at least configure the message to say "sign-out" instead of "sign-in"?

lhuria94:
Thanks @Widcket 

Add the `prompt: login` worked for me. But I see later it can come back to us in case of SSO functionality since we are not clearing the session and making the user to force login again. 

[...]

PR #305, opened 2020-05-19
Added support for ephemeral sessions [SDK-1412]

Widcket:
### Changes

- Added a new boolean parameter called `ephemeralSession` to the `options` object that sets `prefersEphemeralWebBrowserSession` to `true` on iOS 13+. It has no effect on older versions of iOS, or on Android. **Usage of this method disables SSO.**

### References

Closes https://github.com/auth0/react-native-auth0/issues/294

### Testing

This change was tested manually by performing login and logout on iOS and Android.

### Checklist

### DIFF
--- a/src/webauth/__tests__/agent.spec.js
+++ b/src/webauth/__tests__/agent.spec.js
@@ -1,6 +1,6 @@
 jest.mock('react-native');
 import Agent from '../agent';
-import {NativeModules, Linking} from 'react-native';
+import {NativeModules, Linking, Platform} from 'react-native';
 const A0Auth0 = NativeModules.A0Auth0;
 
 describe('Agent', () => {
@@ -37,6 +37,22 @@ describe('Agent', () => {
         const url = 'https://auth0.com/authorize';
         await agent.show(url);
         expect(A0Auth0.url).toEqual(url);
+      });
+
+      it('should not use ephemeral session by default', async () => {
+        Platform.OS = 'ios';
+        expect.assertions(1);
+        const url = 'https://auth0.com';
+        await agent.show(url);
+        expect(A0Auth0.ephemeralSession).toEqual(false);
+      });
+
+      it('should set ephemeral session', async () => {
+        Platform.OS = 'ios';
+        expect.assertions(1);
+        const url = 'https://auth0.com';
+        await agent.show(url, true);
+        expect(A0Auth0.ephemeralSession).toEqual(true);
       });
     });
 
--- a/src/webauth/__tests__/agent.spec.js
+++ b/src/webauth/__tests__/agent.spec.js
@@ -1,6 +1,6 @@
 jest.mock('react-native');
 import Agent from '../agent';
-import { NativeModules, Linking } from 'react-native';
+import {NativeModules, Linking} from 'react-native';
 const A0Auth0 = NativeModules.A0Auth0;
 
 describe('Agent', () => {
@@ -21,14 +21,14 @@ describe('Agent', () => {
     describe('complete web flow', () => {
       beforeEach(() => {
         A0Auth0.onUrl = () => {
-          Linking.emitter.emit('url', { url: 'https://auth0.com' });
+          Linking.emitter.emit('url', {url: 'https://auth0.com'});
         };
       });
 
       it('should resolve promise with url result', async () => {
         expect.assertions(1);
         await expect(
-          agent.show('https://auth0.com')
+          agent.show('https://auth0.com'),
         ).resolves.toMatchSnapshot();
       });
 
@@ -50,7 +50,7 @@ describe('Agent', () => {
 
       it('should remove url listeners when done', async () => {
         A0Auth0.onUrl = () => {
-          Linking.emitter.emit('url', { url: 'https://auth0.com' });
+          Linking.emitter.emit('url', {url: 'https://auth0.com'});
         };
         expect.assertions(1);
         const url = 'https://auth0.com/authorize';
@@ -69,7 +69,7 @@ describe('Agent', () => {
       it('should remove url listeners on first load', async () => {
         expect.assertions(1);
         const url = 'https://auth0.com/authorize';
-        await agent.show(url, true);
+        await agent.show(url, false, true);
         expect(Linking.emitter.listenerCount('url')).toEqual(0);
       });
     });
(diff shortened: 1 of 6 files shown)


src/webauth/agent.js
```diff
@@ -1,12 +1,12 @@
-import { NativeModules, Linking } from 'react-native';
+import {NativeModules, Linking, Platform} from 'react-native';
 
 export default class Agent {
-  show(url, closeOnLoad = false) {
+  show(url, ephemeralSession = false, closeOnLoad = false) {
```
lbalmaceda, line 4: This agent class is public and I think a change like the one introduced here should be considered breaking. 

You are adding a parameter in the middle, those calling today with `show("URL", true) would have a different outcome. Please move it to the end and keep the defaults.

If instead, you consider this class as internal, let's document that and use a "configuration object" instead to make it future proof.

src/webauth/index.js
```diff
@@ -41,15 +41,16 @@ export default class WebAuth {
...
+   * @param {String}  [parameters.nonce] Random string to prevent replay attacks of id_tokens.
+   * @param {String}  [parameters.audience] Identifier of Resource Server (RS) to be included as audience (aud claim) of the issued access token
+   * @param {String}  [parameters.scope] Scopes requested for the issued tokens. e.g. `openid profile`
+   * @param {String}  [parameters.connection] The name of the identity provider to use, e.g. "google-oauth2" or "facebook". When not set, it will display Auth0's Universal Login Page.
+   * @param {Number}  [parameters.max_age] The allowable elapsed time in seconds since the last time the user was authenticated (optional).
+   * @param {Object}  options Other configuration options.
+   * @param {Number}  [options.leeway] The amount of leeway, in seconds, to accommodate potential clock skew when validating an ID token's claims. Defaults to 60 seconds if not specified.
+   * @param {Boolean} [options.ephemeralSession] Disable SSO on iOS 13+. Has no effect on older versions of iOS.
```
lbalmaceda, line 53: I'd reword it to tell that only affects iOS and not Android. Use something like this.

```suggestion
   * @param {Boolean} [options.ephemeralSession] Disable Single-Sign-On (SSO). It only affects iOS with versions 13 and above.
```

src/webauth/agent.js
Widcket: I consider that class to be internal. How would a configuration object future-proof an internal class?

lbalmaceda: My point is that if we keep adding parameters this signature won't look nice. It's OK because it's internal 👍 

A future-proof approach, that doesn't require changing the signature, would be:

```js
show(url, options = {})
const closeOnLoad = options.closeOnLoad || false
const ephemerealSession = options.ephemerealSession || false
```

src/webauth/agent.js
```diff
@@ -1,12 +1,12 @@
...
     if (!NativeModules.A0Auth0) {
       return Promise.reject(
         new Error(
-          'Missing NativeModule. React Native versions 0.60 and up perform auto-linking. Please see https://github.com/react-native-community/cli/blob/master/docs/autolinking.md.'
-        )
+          'Missing NativeModule. React Native versions 0.60 and up perform auto-linking. Please see https://github.com/react-native-community/cli/blob/master/docs/autolinking.md.',
```
lbalmaceda, line 8: I think this extra comma was unintended 
```suggestion
          'Missing NativeModule. React Native versions 0.60 and up perform auto-linking. Please see https://github.com/react-native-community/cli/blob/master/docs/autolinking.md.'
```

src/webauth/index.js
```diff
@@ -41,15 +41,16 @@ export default class WebAuth {
...
+   * @param {String}  [parameters.nonce] Random string to prevent replay attacks of id_tokens.
+   * @param {String}  [parameters.audience] Identifier of Resource Server (RS) to be included as audience (aud claim) of the issued access token
+   * @param {String}  [parameters.scope] Scopes requested for the issued tokens. e.g. `openid profile`
+   * @param {String}  [parameters.connection] The name of the identity provider to use, e.g. "google-oauth2" or "facebook". When not set, it will display Auth0's Universal Login Page.
+   * @param {Number}  [parameters.max_age] The allowable elapsed time in seconds since the last time the user was authenticated (optional).
+   * @param {Object}  options Other configuration options.
+   * @param {Number}  [options.leeway] The amount of leeway, in seconds, to accommodate potential clock skew when validating an ID token's claims. Defaults to 60 seconds if not specified.
+   * @param {Boolean} [options.ephemeralSession] Disable Single-Sign-On (SSO). It only affects iOS with versions 13 and above.
```
lbalmaceda, line 53: Thinking of the public API, and that "ephemeral session" is a concept that Apple introduced for their browser. Should we rename this to something more generic like "disableSSO" in case Android platform support it as well soon? We can keep a reference to ephemeral on the param doc/description

src/webauth/__tests__/agent.spec.js
```diff
@@ -38,6 +38,22 @@ describe('Agent', () => {
         await agent.show(url);
         expect(A0Auth0.url).toEqual(url);
       });
+
+      it('should not use ephemeral session by default', async () => {
+        Platform.OS = 'ios';
```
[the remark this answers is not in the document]
Widcket: Good catch. Added in https://github.com/auth0/react-native-auth0/pull/305/commits/405b4a2c975509ceafcdd39b7c469b87f4e0c9da.

src/webauth/agent.js
Widcket: Shouldn't we remove `closeOnLoad`, anyway? Seems like old leftover code that doesn't do anything now. The native iOS code receives it as a parameter but does nothing with it. Same with the Java code.

src/webauth/index.js
Widcket: Given that "ephemeral session" is not a marketing term (unlike "Universal Links") and that the fact that it disables SSO is just an effect of a broader cause (not storing cookies), it's probably more accurate to leave it that way. It's generic enough IMO.

src/webauth/agent.js
lbalmaceda: iOS uses natively [here](https://github.com/auth0/react-native-auth0/blob/master/ios/A0Auth0.m#L216-L220). 

A side effect of the parameter is to know when to return a value. It's checked [here](https://github.com/auth0/react-native-auth0/blob/master/src/webauth/agent.js#L24-L28). From what I can tell, that's only in use for [authentication](https://github.com/auth0/react-native-auth0/blob/master/src/webauth/index.js#L72) and for [logout](https://github.com/auth0/react-native-auth0/blob/master/src/webauth/index.js#L131).

Widcket: Good idea, but ATM it is just 3 parameters, which is fine IMO. If it keeps growing converting it to an options object is trivial.

PR #313, opened 2020-06-05
Bump dependencies in the lock file

lbalmaceda:
### Changes
- Bumps the dependency versions respecting the package.json file contents.
- Changes the CI script Node version from 8 to LTS

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8
+      - image: circleci/node:lts
     steps:
       - checkout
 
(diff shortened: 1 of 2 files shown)


lbalmaceda:
Getting the following error on the CI
```
error mkdirp@1.0.4: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
error Found incompatible module.
```

I've traced the `mkdirp > 1.0.0` requirement to:
```
jsdoc-to-markdown -> dmd -> cache-point ->mkdirp
```
which we require here as dev dependency. https://github.com/auth0/react-native-auth0/blob/master/package.json#L40

I'll update the package.json file to specify the min engine required, and if required, the CI script to run tests with Node 10

PR #285, opened 2020-03-06
handle invalid token error

emiliokyp:
### Changes

Updated ```AuthError``` class to return invalid_token error from www-authenticate header. Error caused when using an expired token.

### References

issue #241  

### Testing

Manual testing only

- [ ] This change adds unit test coverage
- [x] This change has been tested on the latest version of the platform/language or why not

### Checklist

- [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [x] All existing and new tests complete without errors
- [x] All active GitHub checks have passed

This was just a quick fix to get it working for my project, happy to take feedback and change my approach to this.

PR #329, opened 2020-08-10
Fix the www-authenticate header parsing logic

lbalmaceda:
### Changes

The header object is a special list that must be accessed with `get(name)`, otherwise, there are certain `cors` [scenarios](https://developers.google.com/web/updates/2015/03/introduction-to-fetch#response_types) where not all the headers would be exposed.

This PR handles the possibility of that header not being set and ensures the last fallback is also test-covered.

### References

https://github.com/auth0/react-native-auth0/pull/285 introduced a change that wasn't properly tested

### Testing

### Checklist

### DIFF
--- a/src/auth/__tests__/__snapshots__/index.spec.js.snap
+++ b/src/auth/__tests__/__snapshots__/index.spec.js.snap
@@ -353,9 +353,13 @@ Array [
 ]
 `;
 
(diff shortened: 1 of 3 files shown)


nicosabena:
Awesome, thanks Luciano!

PR #350, opened 2020-12-28
Fixed authentication restart when the app is minimized [SDK-2199]

Widcket:
### Changes

The authentication with the SDK breaks when the screen is minimized and the application is resumed from the application icon. This PR fixes this bug by decoupling the authentication into two activities:
- `AuthenticationActivity`: encapsulates the logic to launch the Custom Tabs intent, and pass the result back.
- `RedirectActivity`: receives the callback URL (via an intent filter), clears the Custom Tabs activity from the stack and passes the callback URL to the `AuthenticationActivity`.

Screen Recording 2020-12-27 at 22 44 45

Both Activity definitions are provided by the SDK. Since the `RedirectActivity` uses intent placeholders, the app must provide the values in the app's `build.gradle` file:

```groovy
android {
    defaultConfig {
        // Add the next line
        manifestPlaceholders = [auth0Domain: "YOUR_AUTH0_DOMAIN", auth0Scheme: "${applicationId}"]
    }
    ...
}
```

**THIS IS A BREAKING CHANGE.**

### Testing

This change was tested manually on an emulator running Android 7.1.1.

### Checklist


android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -52,16 +54,15 @@ public String getName() {
...
-            CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder();
-            CustomTabsIntent customTabsIntent = builder.build();
-            customTabsIntent.launchUrl(activity, Uri.parse(url));
+            AuthenticationActivity.authenticateUsingBrowser(activity, parsedUrl);
         } else {
             final Intent intent = new Intent(Intent.ACTION_VIEW);
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-            intent.setData(Uri.parse(url));
+            intent.setData(parsedUrl);
             getReactApplicationContext().startActivity(intent);
```
lbalmaceda, line 66: how can we ensure an "activity context" is always present? would this branch still work or have the same bug?

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -122,29 +123,35 @@ String generateCodeChallenge(@NonNull String codeVerifier) {
     }
 
     @Override
-    public void onHostResume() {
-        new Handler().postDelayed(new Runnable() {
+    public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
+        new Handler().post(new Runnable() {
```
lbalmaceda, line 127: we probably can get rid of this line. There shouldn't be a need to wrap this on a runnable now that we don't need to post-delayed something.

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -122,29 +123,35 @@ String generateCodeChallenge(@NonNull String codeVerifier) {
...
             @Override
             public void run() {
                 Callback cb = A0Auth0Module.this.callback;
+
                 if (cb != null) {
-                    final WritableMap error = Arguments.createMap();
-                    error.putString("error", "a0.session.user_cancelled");
-                    error.putString("error_description", "User cancelled the Auth");
-                    cb.invoke(error);
+                    if (resultCode == RESULT_OK &&
```
lbalmaceda, line 133: can we extract the 3 conditions into a single variable/flag with a self-describing name?

lbalmaceda (changes requested):
Changes look good, a few questions to resolve. Also, let's add a dedicated section on the top of the README with guidance on how to set up the SDK with these changes. If I'm using the previous version, what changes do I need to make to support this? Is it optional or am I forced to use and declare these new activities in the android manifest?

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Isn't that betond the scope of this PR?

lbalmaceda: Is not out of scope if the other branch still has this bug. Was that verified?

Widcket: But that branch just opens whatever browser the device has installed. That's another process.
Should I check that anyway?

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Done in https://github.com/auth0/react-native-auth0/pull/350/commits/4b7998584f638b1b0909a90a37f7a57967121b53.

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Done in https://github.com/auth0/react-native-auth0/pull/350/commits/4b7998584f638b1b0909a90a37f7a57967121b53.

android/src/main/java/com/auth0/react/A0Auth0Module.java
lbalmaceda: Let me rephrase my original question. If an "Activity context" is not present (activity == null), and so the "else" branch is executed. Would the "authentication restart" bug still occur? and how was that verified?

I'm not familiar with the reasons why the activity is being null-checked. It was [added 4y ago](https://github.com/auth0/react-native-auth0/commit/b841aeb9e1338da6bfee3a6f54ec475d8e3d7c78). I suppose devs were calling the `showUrl()` method from places other than an Activity and this React `getCurrentActivity()` method failed to find one. If having an activity is a requirement for the code, we should document that in the usage/instructions.

Other than using an Activity or an Application as the context, seems like both will end up executing an ACTION_VIEW intent passing the URL as data. Because [CustomTabsIntent.launchUrl(context, url)](https://developers.google.com/web/android/custom-tabs/implementation-guide#what_happens_if_the_user_doesn%E2%80%99t_have_a_browser_that_supports_custom_tabs_installed) will default to that same behavior according to the docs. Does it make sense to maintain 2 separate if/else branches when it can be reduced to just delegating the authentication to the new `AuthenticationActivity`?

Widcket: Indeed, the code needs an Activity to work. I'll add a couple of lines to the Readme.
The `@react-native-community/google-signin` package [straight up produces an error](https://github.com/react-native-google-signin/google-signin/blob/master/android/src/main/java/co/apptailor/googlesignin/RNGoogleSigninModule.java#L189) if the Activity is `null`. And the Activity will be `null` when [there's no Activity attached](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactContext.java#L422) to the current `ReactContext`. And Given that CustomTabs will fallback to a regular browser, it makes no sense to keep the else branch. I'll remove it.

lbalmaceda: > And given that CustomTabs will fallback to a regular browser, it makes no sense to keep the else branch. I'll remove it

👍 

> the code needs an Activity to work.

Let's make this clear in the usage section of the README. Also, if we fail like the linked package above does, let's use a custom error code as we do in other places (e.g. "a0.internal.invalid_android_context") so the developers can match against this code and handle the error better.

lbalmaceda:
@Widcket skipping this PR from 2.8.0 release's contents as this one contains a breaking change. I'll move it to vNext

nlindroos:
Hi! Any chance to expedite getting this to an official version of react-native-auth0? It is a required fix to a critical issue, at least in the way we implement authentication on Android. At the moment we're thus forced to lock react-native-auth0 to a certain commit instead of the official package version to have this fix included.

Widcket:
@nlindroos I expect to pick this up next week, because we'll be changing the implementation a bit.

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Done in https://github.com/auth0/react-native-auth0/pull/350/commits/68b09f3d0360732eb71d7f0bf72fdf4ecff21b3d.

Widcket:
@lbalmaceda the PR has been updated to use intent placeholders. The README has been updated as well, please check the messaging.

https://user-images.githubusercontent.com/5055789/122302581-ac113a00-ced8-11eb-987c-e18f46489e37.mov

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -53,26 +54,23 @@ public String getName() {
...
-            if (activity != null) {
-                CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder();
-                CustomTabsIntent customTabsIntent = builder.build();
-                customTabsIntent.launchUrl(activity, Uri.parse(url));
-            } else {
-                final Intent intent = new Intent(Intent.ACTION_VIEW);
-                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-                intent.setData(Uri.parse(url));
-                getReactApplicationContext().startActivity(intent);
+            if (activity != null) AuthenticationActivity.authenticateUsingBrowser(activity, parsedUrl);
```
lbalmaceda, line 61: let's be consistent and add the braces `{ }` here, please

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -53,26 +54,23 @@ public String getName() {
...
+            else {
+                final WritableMap error = Arguments.createMap();
+                error.putString("error", "a0.activity_not_available");
+                error.putString("error_description", "Android Activity is null.");
```
lbalmaceda, line 65: I think it's a failure scenario in which the dev cannot do anything to recover from it after receiving the error in the callback. Did you run into this issue during your tests?  Would it make sense to just log the error and ignore launching the authentication instead?

The "browser app missing" error is already handled below, that's the important one. Thoughts?

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -131,29 +129,28 @@ String generateCodeChallenge(@NonNull String codeVerifier) {
...
-                    final WritableMap error = Arguments.createMap();
-                    error.putString("error", "a0.session.user_cancelled");
-                    error.putString("error_description", "User cancelled the Auth");
-                    cb.invoke(error);
-                    A0Auth0Module.this.callback = null;
-                }
+    public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
+        Callback cb = A0Auth0Module.this.callback;
+
+        if (cb != null) {
```
lbalmaceda, line 135: I'd flip this condition so we avoid indenting a big part of the code. Something like this is preferred:

if (cb == null){
return;
}

// the rest of the code here

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -131,29 +129,28 @@ String generateCodeChallenge(@NonNull String codeVerifier) {
...
+            boolean hasResult = resultCode == RESULT_OK && 
+                requestCode == AuthenticationActivity.AUTHENTICATION_REQUEST &&
+                data.getData() != null;
+            if (hasResult) cb.invoke(null, data.getData().toString());
```
lbalmaceda, line 139: add braces please

lbalmaceda (changes requested):
Looks good, I left a few comments. Should we add the before/after snippets for the build.gradle and the manifest files, maybe using _diff_ syntax, to highlight the changes required after this breaking change?

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: This is not an error to recover from, it's an error to run into during development and then take the appropriate measures to prevent it (e.g. by making sure there is an activity). I think failing silently is not a desirable outcome.

Widcket: See https://github.com/react-native-google-signin/google-signin/blob/master/android/src/main/java/co/apptailor/googlesignin/RNGoogleSigninModule.java#L189 for a similar example.

Widcket: By making it an error, the dev can send it to Crashlytics if it ever happens on production for some reason.

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Added in https://github.com/auth0/react-native-auth0/pull/350/commits/b5e827bce2820ffb8930272c044f1ccf12e5091e

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Added in https://github.com/auth0/react-native-auth0/pull/350/commits/b5e827bce2820ffb8930272c044f1ccf12e5091e

android/src/main/java/com/auth0/react/A0Auth0Module.java
Widcket: Changed in https://github.com/auth0/react-native-auth0/pull/350/commits/b5e827bce2820ffb8930272c044f1ccf12e5091e

Widcket:
> Should we add the before/after snippets for the build.gradle and the manifest files, maybe using diff syntax, to highlight the changes required after this breaking change?
I can point out to the actual commit with the changes on the RN app, but that would need to  be after the SDK is released. So:
release SDK -> update app -> update SDK README pointing to commit with the changes.

android/src/main/java/com/auth0/react/A0Auth0Module.java
lbalmaceda: Yeah, I raised it because we require an "activity context", and other types of contexts such as an "application context" wouldn't work. We need an activity to launch another activity and await a result from it.

Thanks for the link, is interesting to see that other devs also do it. Let's keep it then.

lbalmaceda (approved):

Bardiamist:
What if I want to continue to use `singleTask` because other dependencies may require it?

Bardiamist:
Can `auth0Domain` store not in `android/app/build.gradle`? Because different environments should use different `auth0Domain` and to replace full `android/app/build.gradle` isn't good. Need a separate config-file.

Widcket:
Hi @Bardiamist, you can add a [string resource file](https://developer.android.com/guide/topics/resources/string-resource) and reference the domain value like this: `auth0Domain: "@string/com_auth0_domain"`.

If you keep the `singleTask`, the auth wil restart when the app is minimized:

https://user-images.githubusercontent.com/5055789/123360612-fd0dd780-d543-11eb-9c4a-8d1ce78010ef.mov

PR #351, opened 2020-12-28
Added support for using a custom scheme in the callback URL [SDK-2223]

Widcket:
### Changes

The Bundle Identifier / Application ID might contain characters not allowed in a URL scheme. This PR adds support for specifying a custom scheme to be used instead.

### Testing

### Checklist

### DIFF
--- a/src/webauth/index.js
+++ b/src/webauth/index.js
@@ -57,6 +57,7 @@ export default class WebAuth {
    * @param {Object}  options Other configuration options.
    * @param {Number}  [options.leeway] The amount of leeway, in seconds, to accommodate potential clock skew when validating an ID token's claims. Defaults to 60 seconds if not specified.
    * @param {Boolean} [options.ephemeralSession] Disable Single-Sign-On (SSO). It only affects iOS with versions 13 and above.
+   * @param {String}  [options.customScheme] Custom scheme to build the callback URL with.
    * @returns {Promise}
    * @see https://auth0.com/docs/api/authentication#authorize-client
    *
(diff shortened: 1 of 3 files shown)


src/webauth/index.js
```diff
@@ -7,15 +7,17 @@ import verifyToken from '../jwt';
 
 const {A0Auth0} = NativeModules;
 
-const callbackUri = domain => {
+const callbackUri = (domain, customScheme) => {
   const bundleIdentifier = A0Auth0.bundleIdentifier;
   const lowerCasedIdentifier = bundleIdentifier.toLowerCase();
-  if (bundleIdentifier !== lowerCasedIdentifier) {
+  if (!customScheme && bundleIdentifier !== lowerCasedIdentifier) {
```
lbalmaceda, line 13: I'd not add this condition here. The reason being that when a custom scheme is in use, the case check for the bundle identifier (used not just in the scheme but also in the path) will not be done. And IIRC every callback URL is case sensitive. I think the warning is still helpful to point out possible casing issues. WDYT?

lbalmaceda (changes requested):
Please add some docs on the method and/or in the README to explain how this can be used and in what scenarios it could be useful.

src/webauth/index.js
Widcket: That's intentional. The bundle identifier is already defined and the developer can't change it, but the custom scheme can be anything the developer wants it to be. So a forced case change is pointless, the developer should be able to specify a valid scheme from the get-go.

Widcket:
> Please add some docs on the method and/or in the README to explain how this can be used and in what scenarios it could be useful.

@lbalmaceda done.

lbalmaceda:
@Widcket Thanks for documenting the parameter. But is not just passing the new option, one would also need to update the intent-filter in the AndroidManifest.xml file, right? 
Developers could benefit from reading that guidance somewhere, e.g. in the readme. I was thinking of adding a small section [like we do here](https://github.com/auth0/react-native-auth0#disable-single-sign-on-ios-13-only) for iOS' SSO. WDYT?

lbalmaceda (approved):

Issue #392, opened 2021-07-19
Auth0 - React Native - Android (release app) - weird behaviour of webAuth Login / logout 


ZubairKh:
Hey everyone!

We have a live app that our users are using! Please need your attention on this problem as quick as possible.

### The Problem 

We have implemented auth0 authentication login into our mobile app built using React Native. We're using webAuth to login (that opens up the webview by auth0 and shows the login form). Also, we did register 2 apps on the auth0 dashboard, one is for our development environment or for testing purposes, another one is for production releaase. 

Our apps are already published on Play and Apple stores and everything seems working fine on both Android and iOS, we were using **`react-native-auth0`** version **`2.6.0`** then we did try to update our app to Play and Apple stores. While making the .aab for Android we were getting the error Manifest merger failed and details error logs were saying we should add **`manifestPlaceholders []`** in build.gradle(app) file. After adding those our build got successful and went live on both the stores. 

Then our QAs noticed on the Android side the login webAuth behaving weirdly like when tapping on the Login button our code was like 

```
const auth0 = new Auth0({
  domain: our_app_name.eu.auth0.com,
  clientId: client_id_from_auth0_dashboard,
})
auth0.webAuth
        .authorize(
          { scope: 'openid email profile' },
          {ephemeralSession: true}
        ).then((credentials) => {
          auth0.auth
            .userInfo({ token: credentials.accessToken })
            .then((data) => {
             // saving data and navigating to Home Page
            })
            .catch((error) => {
              // error logs...
            })
        })
        .catch((error) => {
          // error logs...
        })
```

and we have to tap 3 times in order for this login code to work. On the first tap, we do enter login credentials on webview opened by auth0, then on second tap we see empty webview and disappear after 1 second, third tap same as second and we get a successful response in .then() section of webAuth. The same goes with Logout scenarios like 

 ```
auth0.webAuth
              .clearSession()
              .then((success) => {
                // remove local storage and navigating to Login
              })
              .catch((error) => {
                // error
              })
```

We read the auth0 latest documentation that said 

> Before version 2.9.0, this SDK required you to add an intent filter to the Activity on which you're going to receive the authentication result, and to use the **`singleTask`** **launchMode** in that activity. To migrate your app to version 2.9.0+, remove both and continue with the instructions below.

So we updated the version to 2.9.0, removed intent-filters, and added manifest placeholders as suggested in docs

```
android {
    defaultConfig {
        // Add the next line
[...]

PR #394, opened 2021-07-23
Add a note to the README about manifest placeholders when using flavors

Widcket:
### Changes

This PR adds a note to the README about the need to set up the manifest placeholders differently when using flavors with different application ID values.

### References

From https://github.com/auth0/react-native-auth0/issues/392

### Testing

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ android {
 
 The `applicationId` value will be auto-replaced on runtime with the package name or id of your application (e.g. `com.example.app`). You can change this value from the `build.gradle` file. You can also check it at the top of your `AndroidManifest.xml` file.
 
-> Note that if you're using [flavors](https://developer.android.com/studio/build/build-variants#product-flavors) with different application IDs, you'll need to set up the manifest placeholders for each flavor.
+> Note that if your Android application is using [product flavors](https://developer.android.com/studio/build/build-variants#product-flavors), you might need to specify different manifest placeholders for each flavor.
 
 If you use a value other than `applicationId` in `auth0Scheme` you will also need to pass it as the `customScheme` option parameter of the `authorize` and `clearSession` methods.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +98,8 @@ android {
 
 The `applicationId` value will be auto-replaced on runtime with the package name or id of your application (e.g. `com.example.app`). You can change this value from the `build.gradle` file. You can also check it at the top of your `AndroidManifest.xml` file.
 
+> Note that if you're using [flavors](https://developer.android.com/studio/build/build-variants#product-flavors) with different application IDs, you'll need to set up the manifest placeholders for each flavor.
```
lbalmaceda, line 101: Having different "application IDs" doesn't impact the manifest placeholder contents, as those are used just for domain and scheme.

Furthermore, properties defined in the root level of the `android` block would affect all product flavors. If one requires a different domain, for example, it would need to override it explicitly by defining it again inside their flavor block.

I think the sentence can be improved a bit this way:

```suggestion
> Note that if your Android application is using [product flavors](https://developer.android.com/studio/build/build-variants#product-flavors), you might need to specify different manifest placeholders for each flavor.
```

lbalmaceda (changes requested):
We can add this but IMHO, it feels like an advanced scenario. Anyone working with Android flavours knows these properties in the `build.gradle` file need to either be set globally or overridden per-flavour. Just like when you have different release variants you define signing configuration for each of them.

README.md
Widcket: > Having different "application IDs" doesn't impact the manifest placeholder contents, as those are used just for domain and scheme.

If the app has different app ids defined in each flavor, that definitely impacts the manifest placeholders, as the app id is used as the scheme in the intent filter.

Widcket:
This already tripped up a customer, so it is possible that the omission of this information might cause other customers to make a similar mistake in the future.

However, we can't possibly include every scenario in the Readme. Feel free to close this PR if you think it's not necessary.

lbalmaceda (approved):

PR #455, opened 2022-02-03
Fix iOS module import for Expo SDK 44

Bardiamist:
>From https://github.com/expo/expo/issues/15622#issuecomment-997225774, the double-quoted react header imports will break building on Expo SDK 44. The double-quoted imports should be for react-native < 0.40. as nowadays react-native uses CocoaPods, it should be safe to use the angle-bracket import.

### DIFF
--- a/ios/A0Auth0.h
+++ b/ios/A0Auth0.h
@@ -1,8 +1,4 @@
-#if __has_include("RCTBridgeModule.h")
-#import "RCTBridgeModule.h"
-#else
 #import <React/RCTBridgeModule.h>
-#endif
 
 @interface A0Auth0 : NSObject <RCTBridgeModule>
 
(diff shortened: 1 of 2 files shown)


codebutler:
@evansims How can we help get this merged and released?

poovamraj:
Hey all, We are seriously considering supporting Expo in our project. We will look into this PR soon but it requires some internal discussion that we have already started. I will keep you updated on this and try to provide an update soon. Sorry for the delay!

Widcket (approved):

stevewillard:
Oh my - if this could make it into a new `react-native-auth0` release, that would be amazing! ❤️ 🙏

Bardiamist:
This fix for bare react native projects with Expo. Not for Expo projects.

Issue #460, opened 2022-02-24
On iOS the promise returned by webauth.authorize can sometimes never complete.


marcelmeulemans:
### Describe the problem

The promise returned by `webAuth.authorize()` will consistently not complete when called immediately again after pressing cancel on the iOS permission popup (i.e. "App" want to use...). The issue is related to this `if` statement:

https://github.com/auth0/react-native-auth0/blob/master/src/webauth/agent.js#L44

In reproduction situation both `redirectURL` and `closeOnLoad` are both falsy and the urlHandler will never be called so the promise being returned never completes.

### What was the expected behavior?

The promise should always resolve or reject.

### Reproduction

I can across this behaviour by automatically calling authorize again when the authorize promise rejects. Something like:

```
try {
  await webAuth.authorize();
} catch (_) {
  setTimeout(() => webAuth.authorize(), 0);
}
```
1. Call webAuth.authorize()
2. Press cancel when iOS ask permission for "App" Wants to use "auth0.com" to sign in.
3. Immediately call webAuth.authorize() again.

The returned promise from step 3 will never complete. When the timeout is increased to 1000 the problem does not manifest.

I can probably provide a reproduction scenario via the sample app if the above does not highlight the problem clearly enough.

### Environment

- **Version of this library used:**
  2.13.0
- **Which framework are you using, if applicable:**
  N/A
- **Other modules/plugins/libraries that might be involved:**
  N/A
- **Any other relevant information you think would be useful:**
  `ephemeralSession` is false

Widcket:
Hi @marcelmeulemans, thanks for raising this.
I'm a bit busy ATM, but I'll try to take a look at this next week.

marcelmeulemans:
Thanks @Widcket!

sophie-li:
Thanks so much! Do we have an ETA for when will this be released? 😄

Widcket:
@sophie-li it's out now: https://github.com/auth0/react-native-auth0/releases/tag/v2.13.1

PR #464, opened 2022-03-18
Fix promise that never completes [SDK-3216]

Widcket:
### Changes

When using Web Auth, after the native code calls the RN callback with an error or the redirect URL, the javascript code checks those values and either rejects or resolves a promise. But the code was not handling the case when both error and redirect URL are null, which can happen on iOS under particular circumstances (e.g. retrying the Web Auth login immediately after it failed).

This PR handles that case by rejecting the promise with an error.

### References

Fixes https://github.com/auth0/react-native-auth0/issues/460

### Testing

A unit test was added for this scenario, and also it was tested manually on iOS 15.4 (iPhone simulator), using React Native `0.66.4`.

### Checklist

- [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [ ] All existing and new tests complete without errors

### DIFF
--- a/src/webauth/__tests__/agent.spec.js
+++ b/src/webauth/__tests__/agent.spec.js
@@ -105,6 +105,20 @@ describe('Agent', () => {
         A0Auth0.error = new Error('failed to load');
         await expect(agent.show('https://auth0.com')).rejects.toMatchSnapshot();
       });
+
+      it('should reject with error when both error and redirectURL are missing', async () => {
+        NativeModules.A0Auth0.showUrl = (...args) => {
+          const callback = args[args.length - 1];
+          callback(null, null);
+        };
+        expect.assertions(1);
+        try {
+          await agent.show('https://auth0.com');
+          fail('should have thrown error');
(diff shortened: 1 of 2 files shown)


src/webauth/__tests__/agent.spec.js
```diff
@@ -105,6 +105,20 @@ describe('Agent', () => {
         A0Auth0.error = new Error('failed to load');
         await expect(agent.show('https://auth0.com')).rejects.toMatchSnapshot();
       });
+
+      it('should reject with error when both error and redirectURL are missing', async () => {
+        NativeModules.A0Auth0.showUrl = (...args) => {
```
Widcket, line 110: `NativeModules.A0Auth0` gets reset on every test: https://github.com/auth0/react-native-auth0/blob/master/src/webauth/__tests__/agent.spec.js#L11

poovamraj (approved):
👍 LGTM

Issue #531, opened 2022-10-14
"open with" confirmation on Android after upgrade version


malgorzata-kozera:
### Describe the problem

I have upgraded react-native-auth0 from version 2.12.0 to v2.14.0.
After the upgrade when logging in or logout from the app, the pop-up with the "open with" message appears. The app request confirmation on which app to use to continue (it shows two of the same app icons). There is only one app installed on the device.
It happens on Android.

I followed all recommendations mentioned in these cases:
https://community.auth0.com/t/react-native-auth0-open-with-duplicate-after-upgrade-to-2-9-0/68033
and 
https://github.com/auth0/react-native-auth0/issues/519 
Also, I have only one intent filter in `AndroidManifest.xml` and I removed ` android:launchMode="singleTask"` but it didn't help.

Any solutions for that?

### What was the expected behavior?

- Pop-up for android not shows

### Environment
- react-native-auth0 v2.14.0
- react-native version 0.70.2

poovamraj:
Hi @malgorzata-kozera Thanks a lot for raising this. You are right and we had to fix this on our end in the latest version of our SDK.

This is because we added our Auth0.Android library to provide Credentials Manager support. This added libraries that receive intent for Web Authentication into the RNA library which is why there are 2 intent receivers.

In this [PR](https://github.com/auth0/react-native-auth0/pull/533) we have removed the additional ones from Android which should fix this issue.

Thanks again for raising this 🙌

poovamraj:
We have merged this PR and we will be releasing the fixed version soon. I will update once it is available here. We will close this issue for now but please feel free to reopen it if you have more doubts

poovamraj:
Fix should be available here - https://github.com/auth0/react-native-auth0/releases/tag/v2.14.1

PR #533, opened 2022-10-14
Removed Activity from Android library to avoid prompt

poovamraj:
### Changes

We have removed Activities from Auth0.Android library being added to the RNA manifest. This being added caused to open a prompt with asking to choose the sample application twice. 

These activities are noted in the manifest to be removed

### References

There was an issue noting this here - https://github.com/auth0/react-native-auth0/issues/531

### Testing
We have manually tested this.

We used this command `adb shell dumpsys package r` which previously showed 2 activities registered to receive the intent. Now there will be just 1 which will avoid the above issue

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -29,6 +29,14 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name="com.auth0.android.provider.RedirectActivity"
+            tools:node="remove"/>
+            
+        <activity
+            android:name="com.auth0.android.provider.AuthenticationActivity"
+            tools:node="remove"/>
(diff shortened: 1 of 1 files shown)


Widcket (approved):

mailaneel:
@poovamraj  we are still seeing issues with prompt with this change. when checking for registered intents I am seeing on extra `com.auth0.react.RedirectActivity` registered along with our own one

AndonMitev:
Yeah i'm seeing without having any `RedirectActivity` registered.
```
my-app-scheme:
        975ba9a my-app-scheme/.MainActivity
        a64e666 my-app-scheme/com.auth0.react.RedirectActivity
```

PR #551, opened 2022-11-11
Return a unix timestamp for expiresIn

cranberyxl:
### Changes
You can only return JSON primitives over the bridge, the Foundation Date comes back as `null` for this reason.

### Testing

Look at the credentials returned from `CredentialsManager` via the bridge.

```js
const auth = new Auth0(...) 

auth.credentialsManager.getCredentials().then(console.log)
```

### Checklist

### DIFF
--- a/ios/CredentialsManagerBridge.swift
+++ b/ios/CredentialsManagerBridge.swift
@@ -96,7 +96,7 @@ extension Credentials {
             CredentialsManagerBridge.tokenTypeKey: self.tokenType,
             CredentialsManagerBridge.idTokenKey: self.idToken,
             CredentialsManagerBridge.refreshTokenKey: self.refreshToken as Any,
-            CredentialsManagerBridge.expiresInKey: self.expiresIn,
+            CredentialsManagerBridge.expiresInKey: self.expiresIn.timeIntervalSince1970,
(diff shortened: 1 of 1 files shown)


ios/CredentialsManagerBridge.swift
```diff
@@ -96,7 +96,7 @@ extension Credentials {
             CredentialsManagerBridge.tokenTypeKey: self.tokenType,
             CredentialsManagerBridge.idTokenKey: self.idToken,
             CredentialsManagerBridge.refreshTokenKey: self.refreshToken as Any,
-            CredentialsManagerBridge.expiresInKey: self.expiresIn,
+            CredentialsManagerBridge.expiresInKey: self.expiresIn.timeIntervalSince1970,
```
poovamraj, line 99:
```suggestion
            CredentialsManagerBridge.expiresInKey: floor(self.expiresIn.timeIntervalSince1970),
```

Can we floor this value so that it is returned without decimal points?

poovamraj (changes requested):
That's a great catch! I have added a suggestion. Once it is merged, we can add this to the next release. Thanks a lot!

poovamraj (approved):

PR #552, opened 2022-11-14
Return the auth0 client from the hook

cranberyxl:
### Changes

Return the auth0 client form the hook

```js
const Component = () => {
  const { client } = useAuth0();
  
```

### Testing

### Checklist

### DIFF
--- a/src/hooks/__tests__/use-auth0.spec.js
+++ b/src/hooks/__tests__/use-auth0.spec.js
@@ -99,6 +99,15 @@ describe('The useAuth0 hook', () => {
     expect(result.current.user).not.toBeNull();
   });
 
+  it('initializes the client on start up with valid credentials', async () => {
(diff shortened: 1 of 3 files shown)


poovamraj:
Hi @cranberyxl, can you explain more on the use case for this?

poovamraj:
@cranberyxl The issue with exposing the client is then we wouldn't be able to manage the state. Any change in the user status wouldn't be updated through hooks. The `idToken` is currently available through the credentials.

If you are using `passwordRealm` then currently you can't use it through hooks :/

poovamraj:
Hi @cranberyxl, as mentioned in the previous comment, we cannot expose the auth0 client directly without affecting how the state management works. Due to this reason we will close this PR. 

Please feel free to reopen it if you more doubts. 

We really appreciate all your contribution towards the SDK and we highly value your feedback. So please feel free to share any!

PR #557, opened 2022-11-15
Enable custom params for login methods

poovamraj:
### Changes
Allows sending custom parameters in login methods

### References
This request has been raised internally many times. Other SDKs support custom parameters

Please note any links that are not publicly accessible.

### Testing

### DIFF
--- a/src/auth/__tests__/index.spec.js
+++ b/src/auth/__tests__/index.spec.js
@@ -833,6 +833,16 @@ describe('auth', () => {
       expect.assertions(1);
       await expect(auth.loginWithOTP(parameters)).resolves.toMatchSnapshot();
     });
+
+    it('should handle custom parameters', async () => {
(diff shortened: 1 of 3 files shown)


src/auth/__tests__/index.spec.js
```diff
@@ -380,6 +427,19 @@ describe('auth', () => {
...
+        expect.assertions(1);
+        await auth.loginWithSMS({
+          phoneNumber: '+5491159991000',
+          code: '123456',
+          audience: 'http://myapi.com',
+          scope: 'openid',
+          custom: 'should not be filtered',
+        });
+        expect(fetchMock.lastCall()).toMatchSnapshot();
+      });
```
Widcket, line 442: Should we add similar tests for `loginWithOTP()`, and `loginWithOOB()`, `loginWithRecoveryCode()` as well?

Widcket (approved):

PR #562, opened 2022-11-22
Added saveCredentials function to useAuth0 hook

michalpazur:
### Changes

This PR makes `saveCredentials` function from `CredentialManager` accessible via the `useAuth0` hook.

### Testing

This change was tested on iPhone 14 (iOS 16.1) simulator with React Native 0.70.5 and React 18.1.0.

### Checklist

### DIFF
--- a/src/hooks/__tests__/use-auth0.spec.js
+++ b/src/hooks/__tests__/use-auth0.spec.js
@@ -432,4 +432,26 @@ describe('The useAuth0 hook', () => {
     await waitForNextUpdate();
     expect(result.current.error).toEqual(thrownError);
   });
+
+  it('can save new credentials', async () => {
+    const {result, waitForNextUpdate} = renderHook(() => useAuth0(), {wrapper});
+
+    result.current.saveCredentials({parameter: 1});
+    await waitForNextUpdate();
+
+    expect(mockAuth0.credentialsManager.saveCredentials).toHaveBeenCalledWith({
+      parameter: 1,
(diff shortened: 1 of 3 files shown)


poovamraj:
@michalpazur since the credentials are already stored internally always, what scenario will this hook provide? Is this for API client based login and then using hooks from there after storing the credentials?

poovamraj:
@michalpazur Any update on this? We would love to understand how you would like to use the library and help you with this feature. Since there is also some community engagement in this we are keen to hear all your feedback.

poovamraj:
Hi @michalpazur, we are planning to support all authentication methods through hooks and in that case there would be no need to save credentials seperately.. for this reason we would like to close this PR.

Please feel free to reopen this if you find it fit or want to add more information.

evrimfeyyaz:
Hi @poovamraj. I would also like to see this added to the library.

This would be useful for switching between multiple logged-in users.

1. User logs in as "User A," the credentials are saved in Auth0's credentials manager, and they are also stored separately on the device.
2. They would like to log in as "User B," and when they do, Auth0's credentials are replaced by User B's credentials.
3. Now, they would like to go back to using "User A" on the device, so we directly replace the credentials within Auth0's credentials manager with User A's credentials (which were separately stored in Step 1).

I know this is a bit of a convoluted way of doing this, but short of implementing a multi-user credentials manager within Auth0, this seems to be the best option for supporting this use case.

PR #570, opened 2022-12-09
Fixing `.hide()` on iOS

stigi:
Currently calling `.hide()` on iOS only works on pre iOS pre 11.

Pre iOS 11 the library is presenting the authentication webview using `presentingViewController`. On iOS 11 `SFSafariViewController` is used though, an on iOS 12 and up we're using `ASWebAuthenticationSession`.

This PR is implementing the required changes to close the authentication webview on iOS 11 and up.


Widcket (approved):
Thanks for the PR @stigi!

PR #584, opened 2023-01-18
getCredentials hook now updates user state

KMathisGit:
### Changes

`getCredentials` hook method now updates `user` state. This is important because when getting new credentials the ID token may be renewed, if renewed, the ID token's claims may now be different and so the `user` state (which is derived from these claims) should also be updated.

### Testing

If a user's ID token changes in some way (meta data perhaps) and it going to be renewed (depends on `expiresAt` credentials value) then upon calling `getCredentials` - the `user` state should reflect the respective changes made to the ID token.

### Checklist


poovamraj:
Hi @KMathisGit, this is a great and very welcomed PR. We would love to accept this with the following changes.

- Instead of changing the action to `SET_USER`, you can introduce a different action and in the reducer, use same code block for both the cases like our auth0-react library [here](https://github.com/auth0/auth0-react/blob/master/src/reducer.tsx#L28)
- We would like unit tests that would do the following 
    -  user is updated if there is a change
    -  user is not updated if there is no change

poovamraj:
@frederikprijck since I have done some changes to this PR I wanted another set of eyes looking at it before I merge it. Since the changes are closest to `auth0-react` would be helpful to get your review on this

frederikprijck (approved):

Issue #606, opened 2023-03-24
Stuck at log in after entering credentials on Android, doesn't redirect back or log me in


ttlenell:
### Describe the problem

Issues with loggin in with an android device / simulator. After entering the credentials in the browser that appears, I press "Continue" and I see in the top of my browser that it tries to load the request but stops loading after about one second and then nothing happens. Nothing happens if I press the "Continue"-button either. 

But this only happens when I enter the correct credentials. If I enter the wrong ones on purpose, it will give me an error that its wrong. 

And when I press back/exit the browser to return to the app to try again, the browser just stays blank. Nothing happens and I can't trigger anything. 

This ONLY happens on android, iOS works just fine.

### What was the expected behavior?

Be able to log in without issue on android. After I've put in credentials and pressed "Continue", it should redirect me back to app and I should be logged in.

### Reproduction
Set up a new expo project with Typescript and the latest SDK.
Add `react-native-auth0` & `@types/react-native-auth0`.

Copy the example code detailed here for example
https://github.com/auth0-samples/auth0-react-native-sample/tree/master/00-Login-Expo

### Environment

react-native-auth0@2.17.1
@types/react-native-auth0@2.14.1
react-native@0.71.4
expo@48.0.9

react-native info:

```
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1
    Memory: 129.78 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
    Watchman: 2022.08.22.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK:
      API Levels: 23, 26, 28, 29, 30, 31, 33
      Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3, 31.0.0, 33.0.0
      System Images: android-31 | Google APIs ARM 64 v8a
      Android NDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9619390
[...]

Issue #582, opened 2023-01-17
Android hangs in RN 0.70.6


shawna-donnelly:
### Describe the problem
Auth0 working properly for ios, but when I try to run it for android, it hangs after selecting my email address from the Google login. I am not seeing any kind of log other than TrafficStats with tagSocket(80) in LogCat. Otherwise, no feedback at all. 

### What was the expected behavior?
I expect the login to complete. 

### Reproduction

Started a new typescript project, which used 0.70.6. Used useAuth0 and added a provider. Updated manifest placeholders to use my auth0 domain. Ran project, and it is not working. Works properly in iOS. 

### Environment

"react": "18.1.0",
        "react-native": "0.70.6",
        "react-native-auth0": "^2.16.0",

- **Version of this library used:**
- **Which framework are you using, if applicable:**
- **Other modules/plugins/libraries that might be involved:**
- **Any other relevant information you think would be useful:**

shawna-donnelly:
Seems to work on the emulator. Not sure why it doesn't work on my physical device.

poovamraj:
Can you share the device details? Also reproducing this on our [sample repository](https://github.com/auth0-samples/auth0-react-native-sample) would make it a lot easier to debug it.

owentar:
We're having similar issue (probably the same). Will try to add more info.

We've created an expo app and we're trying to introduce `react-native-auth0` to authenticate our users. Following the steps described [here](https://auth0.com/docs/quickstart/native/react-native-expo/00-login).

With all that, we're simply trying to test it first displaying a login button (when the user is not logged in) and a logout button when it is logged in). Pretty much as shown in the [sample repository](https://github.com/auth0-samples/auth0-react-native-sample/blob/master/00-Login-Expo/App.js)

**This works fine in iOS but doesn't work on Android**. We added a `console.log` to inspect the `error` reported by the hook. We can see the following message: `{"error": "a0.session.user_cancelled", "error_description": "User cancelled the Auth"}`.

NOTE: **sometimes seems to be able to login, but most of the times this is not the case and the error can be seen in the console**. Again, this is only happening on Android. iOS seems to be working fine.

Repo reproducing the issue: https://github.com/owentar/rn-auth0-expo-test

@shawna-donnelly please confirm this is the same issue or we need to create a separate one.

poovamraj:
@owentar can you try to reproduce this it in this branch - https://github.com/auth0/react-native-auth0/tree/test/existing-task-redirection-issue?

[...]

PR #610, opened 2023-03-31
Make custom scheme mandatory for Expo

poovamraj:
### Changes

With this documentation change, we are making custom scheme mandatory. This is due to the reason that Expo uses the application ID or bundle identifier we use to as the scheme to open the application. This shows a popup when the login is done successfully and choosing the wrong option could break the application flow.

For this reason we have made custom scheme as mandatory for Expo application and added guidance on how to do it.

This will change in the next major where we will generate the scheme for callback URL with `auth0` prefixed to the package name to avoid clash with Expo

### References

Issues were raised because of this 

https://github.com/auth0/react-native-auth0/issues/582
https://github.com/auth0/react-native-auth0/issues/606

PR in samples - https://github.com/auth0-samples/auth0-react-native-sample/pull/113

### Testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,8 +336,7 @@ const Component = () => {
   const {clearSession, user} = useAuth0();
 
   const logout = async () => {
-    await clearSession();
-    //await clearSession({customScheme: 'CUSTOM_SCHEME'); //When using Expo or custom scheme
+    await clearSession(); // clearSession({customScheme: 'CUSTOM_SCHEME') when using Expo or a custom scheme
   };
 
   return <View>{user && <Button onPress={logout} title="Log out" />}</View>;
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ If you use a value other than `$(PRODUCT_BUNDLE_IDENTIFIER)` in the `CFBundleURL
 
 > :warning: This SDK is not compatible with "Expo Go" app because of custom native code. It is compatible with Custom Dev Client and EAS builds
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -337,6 +338,7 @@ const Component = () => {
 
   const logout = async () => {
     await clearSession();
+    //await clearSession({customScheme: 'CUSTOM_SCHEME'); //When using Expo or custom scheme
```
Widcket, line 341:
```suggestion
    await clearSession(); // clearSession({customScheme: 'CUSTOM_SCHEME') when using Expo or a custom scheme
```

README.md
```diff
@@ -286,6 +286,7 @@ const Component = () => {
 
   const login = async () => {
     await authorize();
+    //await authorize({}, {customScheme: 'CUSTOM_SCHEME'); //When using Expo or custom scheme
```
Widcket, line 289:
```suggestion
    await clearSession(); // clearSession({customScheme: 'CUSTOM_SCHEME') when using Expo or a custom scheme
```

Widcket (dismissed):
>This will change in the next major where we will generate the scheme for callback URL with auth0 prefixed to the package name to avoid clash with Expo

In the next major we could also not use `https` as the default scheme on Android, and let developers opt-in to App Links.

poovamraj:
@Widcket currently RNA doesn't have a place where we use `https` as scheme unless the developer opts in mandatorily using custom scheme. Is that what you are asking?

Issue #340, opened 2020-11-02
android.content.ActivityNotFoundException: No Activity found to handle Intent


ar-g:
We're getting quite a few exceptions on the Android side. I believe that the main cause is that we need to handle a case when the Browser app is not available on a specific Android device. So the solution would be to wrap [this](https://github.com/auth0/react-native-auth0/blob/d6be771267b750058fdd039db7da1ae9b8c73da0/android/src/main/java/com/auth0/react/A0Auth0Module.java#L53) into the try-catch block and propagate the error to ReactNative and ask the user to download the Browser on the library consumer side. Another option is to check in advance if the browser is available https://developers.google.com/web/android/custom-tabs/implementation-guide#what_happens_if_the_user_doesn%E2%80%99t_have_a_browser_that_supports_custom_tabs_installed

Stacktrace:
```
* android.app.Instrumentation.checkStartActivityResult Instrumentation.java:2014

* android.app.Instrumentation.execStartActivity Instrumentation.java:1675

* android.app.Activity.startActivityForResult Activity.java:4601

* androidx.fragment.app.FragmentActivity.startActivityForResult FragmentActivity.java:6

* android.app.Activity.startActivityForResult Activity.java:4559

* androidx.fragment.app.FragmentActivity.startActivityForResult FragmentActivity.java:3

* android.app.Activity.startActivity Activity.java:4920

* androidx.core.content.ContextCompat.startActivity ContextCompat.java:2

* androidx.browser.customtabs.CustomTabsIntent.launchUrl CustomTabsIntent.java:2

* com.auth0.react.A0Auth0Module.showUrl A0Auth0Module.java:5
```

I can create a PR for this to speed things up as this is causes a lot of issues for us.

lbalmaceda:
@ar-g I don't think this exception is for the SDK to handle. In the end, the custom tabs dependency will always fallback to a simple `ACTION_VIEW` intent passing the URL you want to open as data. You could query the package manager for activies that can resolve that action in advance, before launching the web auth. I find hard to believe that nowadays there are still devices without a single browser app installed. but is still valid to check for availability.

ar-g:
👍 That's a fair point, we can do this on the client-side. But since the library fires an intent I thought that would be more appropriate to do it on the library side.

Regarding nowadays devices, we have the Stacktrace:

```
androidx.fragment.app.FragmentActivity.startActivityForResult
FragmentActivity.java, line 6
[...]

PR #611, opened 2023-03-31
[ESD-27178] Fix browser not found issue not being surfaced

poovamraj:
### Changes
Currently when a browser is not available in Android, the issue is not surfaced and instead it gets engulfed. To fix this issue we have added a check to result code to check whether browser is not found.

The error written [here](https://github.com/auth0/react-native-auth0/blob/master/android/src/main/java/com/auth0/react/A0Auth0Module.java#L160) doesn't seem to be raised

### References
https://github.com/auth0/react-native-auth0/issues/340

### Testing


Widcket (commented):
If `authenticateUsingBrowser()` does not throw an `ActivityNotFoundException`, should we remove the try-catch in https://github.com/auth0/react-native-auth0/blob/master/android/src/main/java/com/auth0/react/A0Auth0Module.java#L157? Or is there a case where it does throw?

poovamraj:
@Widcket there are cases in which it throws as well. So we have not removed it.

Widcket (approved):

Issue #614, opened 2023-04-10
return `type` not `tokenType` from getCredentials on Android


sottar:
### Describe the problem

We are calling `credentialsManager.getCredentials()` / `credentialsManager.saveCredentials()` directly from our ReactNative app.   
 `getCredentials` returns `type` instead of `tokenType` on Android. In iOS, `tokenType` is returned as the [type definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native-auth0/index.d.ts#L283).

It seems to depend on an Android specification. 
https://javadoc.io/static/com.auth0.android/auth0/1.14.1/com/auth0/android/result/Credentials.html

Is this the intended specification of react-native-auth0?
If so, we should update the getCredentials type along with it.

### What was the expected behavior?

`getCredentials()` returns tokenType on both of iOS and Andorid

### Reproduction

This is not whole codes, but we call functions like this.

```
import Auth0, { SaveCredentialsParams } from 'react-native-auth0';

export const Auth0Provider: FC<Props> = ({ children }) => {
	const client = new Auth0({
	  domain: 'xxx',
	  clientId: 'yyy',
	})
	
	const saveCredentials = async (credentials: SaveCredentialsParams) => {
	    await client.credentialsManager.saveCredentials(credentials);
	}
	
	const getCredentials = async () => {
	    return await client.credentialsManager.getCredentials(credentials);
	}

	return { saveCredentials, getCredentials }
}

```

### Environment

- **Version of this library used:** 2.17.1
- **Which framework are you using, if applicable:** ReactNative
- **Other modules/plugins/libraries that might be involved:**
	- ReactNative: 0.71.1,
	- Android Emulator: Pixel_3a_API_33_arm46-v8a
- **Any other relevant information you think would be useful:**

poovamraj:
@sottar Thanks for raising this issue. We will add `tokenType` to the credentials object and remove existing `type` from our object in the next major as it could lead to a breaking change.

poovamraj:
Hi @sottar, we are fixing this in https://github.com/auth0/react-native-auth0/pull/621

Please feel free to follow the progress there or to share your thoughts. We will close this issue now.

PR #621, opened 2023-04-26
Added tokenType to credentials returned in Android

poovamraj:
### Changes
Android has been returning `type` in Credentials object instead of `tokenType` so we are adding this. We will be removing `type` in a major release as it could be a breaking change if we remove it now

### References
https://github.com/auth0/react-native-auth0/issues/614

### Testing


android/src/main/java/com/auth0/react/CredentialsParser.java
```diff
@@ -29,7 +29,8 @@ public static ReadableMap toMap(Credentials credentials) {
         map.putString(ID_TOKEN_KEY, credentials.getIdToken());
         map.putString(SCOPE, credentials.getScope());
         map.putString(REFRESH_TOKEN_KEY, credentials.getRefreshToken());
-        map.putString(TYPE_KEY, credentials.getType());
+        map.putString(TYPE_KEY, credentials.getType());//@deprecate This should be removed in next major (Reference - https://github.com/auth0/react-native-auth0/issues/614)
+        map.putString(TOKEN_TYPE_KEY, credentials.getType());
         return map;
```
Widcket, line 34: Should we be updating any tests?

poovamraj: Since this is android specific and we don't have native unit tests, no unit tests need to be updated

poovamraj: Since this is android specific and we don't have native unit tests, no unit tests need to be updated

Widcket (approved):

PR #637, opened 2023-06-09
[SDK-3598] RNA sample app

poovamraj:
### Changes
- Added a sample app to the repo
- Moved React Native to a peer dependency (this caused linking issue in sample app)
- Fixed a parameter error in webauth/index.js
- We are not using workspace feature from yarn as React Native depends on node_module location in particular place which wouldn't work with workspace

### Testing


poovamraj:
We have an internal study on why we are closing this PR. Please refer to it in the ticket number mentioned in the PR title

PR #639, opened 2023-06-13
[SDK-3598] RNA sample app

poovamraj:
### Changes
- Added a sample app to the repo
- Moved React Native to a peer dependency (this caused linking issue in sample app)
- Fixed a parameter error in webauth/index.js
- This is done by using https://github.com/callstack/react-native-builder-bob

### Testing


stevehobbsdev:
@poovamraj How should I build, configure and run the sample? That info would be good to see in the readme as well for others.

poovamraj:
@stevehobbsdev I created a seperate README inside the example folder, this way we can avoid novice users from getting confused because of the example application guide

stevehobbsdev:
I've got a few issues running the sample on iOS according to the instructions:

image

stevehobbsdev (changes requested):
As mentioned on our call to resolve the above (thank you!) I would just tweak the readme:

* Command to run should be `yarn example ios|android`
* Let's see if we can consolidate all the `yarn` commands down to just one or two
* Let's include a short section in the readme on where to go to configure with my own tenant - great that it has a default though

poovamraj:
Hey @stevehobbsdev great feedback. I have implemented them. I have left `bootstrap` and `prepack` as it is instead of merging it as they are lifecycle commands that come inbuilt and I feel it is better to leave them as it is.

poovamraj:
This PR will improve the test coverage - https://github.com/auth0/react-native-auth0/pull/652. As we have done the native code migration unit tests there

stevehobbsdev (approved):

poovamraj:
As mentioned before,  This PR will improve the test coverage - https://github.com/auth0/react-native-auth0/pull/654. As we have done the native code migration unit tests there.

stevehobbsdev (approved):

Issue #646, opened 2023-06-15
[EXPO][iOS]: compiling for iOS 12.0, but module 'Auth0' has a minimum deployment target of iOS 13.0


nickrugge:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

expo: 48
react-native: 0.71.8
react-native-auth0: 2.17.2 to 2.17.4
platform: ios

The fastlane build phase produces the following error:

```
❌  (node_modules/react-native-auth0/ios/CredentialsManagerBridge.swift:8:8)

   6 | //
   7 | 
>  8 | import Auth0
     |        ^ compiling for iOS 12.0, but module 'Auth0' has a minimum deployment target of iOS 13.0: /Users/expo/workingdir/build/ios/build/Build/Products/Debug-iphonesimulator/Auth0/Auth0.swiftmodule/x86_64-apple-ios-simulator.swiftmodule
   9 | import Foundation
  10 | import LocalAuthentication
  11 | 
  ```

We have also tried installing `expo-build-properties` package and configure the deploymentTarget to iOS 13 for the project

```
plugins: [
   [
      'expo-build-properties',
      {
       ios: {
          deploymentTarget: '13.0',
       },
    },
  ],
]
```

but it didn't work even if the project deployment target was set to 13.0.

The build stopped working since yesterday under the exactly same conditions as before when it worked, looking at the build output logs the pods installed are changed.

```
Installing A0Auth0 (2.17.2)
Installing Auth0 (2.3.2)
```
to
```
[...]

PR #647, opened 2023-06-15
Fix Auth0 pod version to constant

poovamraj:
### Changes
Using dynamic pod version to fetch the latest version of the Auth0.Swift SDK seems to break RNA. We have used fixed version because of this

### References
https://github.com/auth0/react-native-auth0/issues/646

### Testing

### DIFF
--- a/A0Auth0.podspec
+++ b/A0Auth0.podspec
@@ -17,5 +17,5 @@ Pod::Spec.new do |s|
   s.requires_arc = true
 
   s.dependency 'React-Core'
-  s.dependency 'Auth0', '~> 2.3'
+  s.dependency 'Auth0', '2.3'
 end


A0Auth0.podspec
```diff
@@ -17,5 +17,5 @@ Pod::Spec.new do |s|
   s.requires_arc = true
 
   s.dependency 'React-Core'
-  s.dependency 'Auth0', '~> 2.3'
+  s.dependency 'Auth0', '2.3'
```
Widcket, line 20:
```suggestion
  s.dependency 'Auth0', '2.3.2'
```
That's the last release that supports iOS 12.

Widcket (approved):

Widcket (approved):

PR #654, opened 2023-06-20
[SDK-4355] iOS Native Code Migration

poovamraj:
### Changes
- Added support for forceRefresh in iOS
- Added support for additional parameters in iOS and Android
- Migrated native code to use Auth0.Swift
- Refactors CredentialsManagerBridge to NativeBridge
- Moved all the required unit tests to native code

### Testing


android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -140,7 +140,13 @@ public String getName() {
     }
 
     @ReactMethod
-    public void webAuth(String scheme, String state, String nonce, String audience, String scope, String connection, int maxAge, String organization, String invitationUrl, int leeway, Promise promise) {
+    public void webAuth(String scheme, String state, String nonce, String audience, String scope, String connection, int maxAge, String organization, String invitationUrl, int leeway, boolean ephemeralSession, ReadableMap additionalParameters, Promise promise) {
```
Widcket, line 143: Is `ephemeralSession` needed in the method signature for the Android platform?

ios/A0Auth0.xcodeproj/project.pbxproj
```diff
@@ -247,6 +248,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
+				EXCLUDED_ARCHS = arm64;
```
Widcket, line 251: Does this mean it won't work on Apple Silicon macs?

ios/A0Auth0.xcodeproj/project.pbxproj
```diff
@@ -226,6 +226,7 @@
...
```
Widcket, line 229: Does this mean it won't work on Apple Silicon macs?

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+        if(ephemeralSession) {
+            let _ = builder.useEphemeralSession()
+        }
+        builder.parameters(additionalParameters)
+        builder.start { result in
+            switch result {
+            case .success(let credentials):
+                resolve(credentials.asDictionary())
+            case .failure(let error):
+                reject(NativeBridge.errorCode, error.errorDescription, error)
```
Widcket, line 77: Won't this use `a0.invalid_state.credential_manager_exception` as the error code for Web Auth login?

Widcket: Note that Web Auth operations produce a `WebAuthError`, which have different cases, according to what the error is: https://auth0.github.io/Auth0.swift/documentation/auth0/webautherror

In the Flutter SDK we map the cases to unique error codes: https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/ios/Classes/WebAuth/WebAuthExtensions.swift#L4

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+    }
+    
+    @objc public func webAuthLogout(scheme: String?, federated: Bool, resolve: @escaping RCTPromiseResolveBlock, reject: @escaping RCTPromiseRejectBlock) {
+        let builder = Auth0.webAuth(clientId: self.clientId, domain: self.domain)
+        builder.clearSession(federated: federated) { result in
+                switch result {
+                case .success:
+                    resolve(true)
+                case .failure(let error):
+                    reject(NativeBridge.errorCode, error.errorDescription, error)
```
Widcket, line 90: Same as above.

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+             expiresIn = dateFormatter.date(from: dateStr)
+         }
+     
+        
+        let credentials =  Credentials(
+            accessToken: accessToken,
+            tokenType: tokenType,
+            idToken:  idToken,
+            refreshToken: refreshToken,
+            expiresIn: expiresIn ?? Date(),
```
Widcket, line 118: If `expiresIn` is `nil`, should it be an error instead of using the current date?

src/credentials-manager/index.ts
```diff
@@ -68,6 +68,7 @@ class CredentialsManager {
    * @param {String} scope optional - the scope to request for the access token. If null is passed, the previous scope will be kept.
    * @param {String} minTtl optional - the minimum time in seconds that the access token should last before expiration.
    * @param {Object} parameters optional - additional parameters to send in the request to refresh expired credentials.
+   * @param {Object} forceRefresh optional - to force refresh the credentials. It will work only if refresh token already exists.
```
Widcket, line 71:
```suggestion
   * @param {Object} forceRefresh optional - to force refresh the credentials. It will work only if the refresh token already exists.
```

ios/A0Auth0.m
```diff
@@ -37,58 +19,43 @@ - (dispatch_queue_t)methodQueue
...
-    if (@available(iOS 11.0, *)) {
-        self.sessionCallback = callback;
-        self.closeOnLoad = closeOnLoad;
-        [self presentAuthenticationSession:[NSURL URLWithString:urlString] usingEphemeralSession:ephemeralSession];
-    } else {
-        [self presentSafariWithURL:[NSURL URLWithString:urlString]];
-        self.sessionCallback = callback;
-        self.closeOnLoad = closeOnLoad;
-    }
+RCT_EXPORT_METHOD(webAuth:(NSString *)scheme state:(NSString *)state nonce:(NSString *)nonce audience:(NSString *)audience scope:(NSString *)scope connection:(NSString *)connection maxAge:(NSInteger)maxAge organization:(NSString *)organization invitationUrl:(NSString *)invitationUrl  leeway:(NSInteger)leeway ephemeralSession:(BOOL)ephemeralSession additionalParameters:(NSDictionary *)additionalParameters resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
```
Widcket, line 53: Is `scheme` needed in the method signature for the iOS platform?

ios/A0Auth0.m
```diff
@@ -37,58 +19,43 @@ - (dispatch_queue_t)methodQueue
...
+    [self.nativeBridge webAuthWithScheme:scheme state:state nonce:nonce audience:audience scope:scope connection:connection maxAge:maxAge organization:organization invitationUrl:invitationUrl leeway:leeway ephemeralSession:ephemeralSession additionalParameters:additionalParameters resolve:resolve reject:reject];
 }
 
-RCT_EXPORT_METHOD(oauthParameters:(RCTResponseSenderBlock)callback) {
-    callback(@[[self generateOAuthParameters]]);
+RCT_EXPORT_METHOD(webAuthLogout:(NSString *)scheme federated:(BOOL)federated resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
```
Widcket, line 57: Is `scheme` needed in the method signature for the iOS platform?

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+    @objc public init(clientId: String, domain: String) {
+        let auth0 = Auth0
+            .authentication(clientId: clientId, domain: domain)
+        self.clientId = clientId
+        self.domain = domain
+        self.credentialsManager = CredentialsManager(authentication: auth0)
+        super.init()
+   }
+    
+    @objc public func webAuth(scheme: String?, state: String?, nonce: String?, audience: String?, scope: String?, connection: String?, maxAge: Int, organization: String?, invitationUrl: String?, leeway: Int, ephemeralSession: Bool, additionalParameters: [String: String], resolve: @escaping RCTPromiseResolveBlock, reject: @escaping RCTPromiseRejectBlock) {
```
Widcket, line 42: Is `scheme` needed here?

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+            case .success(let credentials):
+                resolve(credentials.asDictionary())
+            case .failure(let error):
+                reject(NativeBridge.errorCode, error.errorDescription, error)
+            }
+        }
+            
+    }
+    
+    @objc public func webAuthLogout(scheme: String?, federated: Bool, resolve: @escaping RCTPromiseResolveBlock, reject: @escaping RCTPromiseRejectBlock) {
```
Widcket, line 83: Is `scheme` needed here?

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+    }
+    
+    @objc public func getCredentials(scope: String?, minTTL: Int, parameters: [String: Any], forceRefresh: Bool, resolve: @escaping RCTPromiseResolveBlock, reject: @escaping RCTPromiseRejectBlock) {
+        if(forceRefresh) {
+            credentialsManager.renew { result in
+                switch result {
+                case .success(let credentials):
+                    resolve(credentials.asDictionary())
+                case .failure(let error):
+                    reject(NativeBridge.errorCode, error.errorDescription, error)
```
Widcket, line 132: Note that the Credentials Manager produces `CredentialsManagerError` error values, which have different cases, depending on what the error is: https://auth0.github.io/Auth0.swift/documentation/auth0/credentialsmanagererror

In the Flutter SDK we map the cases to unique error codes: https://github.com/auth0/auth0-flutter/blob/e9dfa49419a1047bfaaddaf7f752fbbed3c400e3/auth0_flutter/ios/Classes/CredentialsManager/CredentialsManagerExtensions.swift#L4

ios/NativeBridge.swift
```diff
@@ -0,0 +1,181 @@
...
+                }
+            }
+        } else {
+            credentialsManager.credentials(withScope: scope, minTTL: minTTL, parameters: parameters) { result in
+                switch result {
+                case .success(let credentials):
+                    resolve(credentials.asDictionary())
+                case .failure(let error):
+                    reject(NativeBridge.errorCode, error.errorDescription, error)
```
Widcket, line 141: Same as above.

src/webauth/index.ts
```diff
@@ -180,33 +80,20 @@ class WebAuth {
...
-      logoutParameters.federated = parameters.federated || false;
-      const logoutUrl = client.logoutUrl(logoutParameters);
-      return agent.show(logoutUrl, false, options.skipLegacyListener, true);
-    }
+  clearSession(parameters: ClearSessionParameters = {}) {
+    const { agent, domain, clientId } = this;
+    return agent.logout(
+      { clientId, domain },
+      {
+        customScheme: parameters.customScheme,
```
Widcket, line 93: `customScheme` is part of the options for `authorize()` (see https://github.com/auth0/react-native-auth0/blob/vnext/src/webauth/index.ts#L77), but here is part of the parameters. Should it be a parameter or an option?

Widcket:
Should we add native unit tests?

android/src/main/java/com/auth0/react/A0Auth0Module.java
poovamraj: The method signature for iOS and Android has to be same. So we are adding the missing parameters as well.

ios/A0Auth0.m
poovamraj: The method signature for iOS and Android has to be same. So we are adding the missing parameters as well.

ios/NativeBridge.swift
poovamraj: The method signature for iOS and Android has to be same. So we are adding the missing parameters as well.

ios/NativeBridge.swift
poovamraj: The method signature for iOS and Android has to be same. So we are adding the missing parameters as well.

src/webauth/index.ts
poovamraj: `agent.logout` has different API than what we expose outside. `agent.login` and `agent.logout` take `clientId` and `domain` as parameters and all the other as options. The external API still remains the same.

ios/A0Auth0.m
poovamraj: The method signature for iOS and Android has to be same. So we are adding the missing parameters as well.

ios/NativeBridge.swift
Widcket: But this is the Swift code, that gets manually called from the ObjC code. Does it still need to have the same signature?

poovamraj:
Unit tests for native code is done for the GA and tracked as part of SDK-4374.

ios/NativeBridge.swift
```diff
@@ -0,0 +1,220 @@
...
+            NativeBridge.idTokenKey: self.idToken,
+            NativeBridge.refreshTokenKey: self.refreshToken as Any,
+            NativeBridge.expiresInKey: floor(self.expiresIn.timeIntervalSince1970),
+            NativeBridge.scopeKey: self.scope as Any
+        ]
+    }
+}
+
+extension WebAuthError {
+    func getReactNativeErrorCode() -> String {
```
Widcket, line 189: The [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) state that methods that do not have side effects should read as nouns:

Screenshot 2023-06-23 at 10 51 17 AM

`getReactNativeErrorCode()` being a pure function, should instead be named something like `reactNativeErrorCode()`. Or simply make it a computed property, since it takes no parameters. In which case it should still read as a noun:

Screenshot 2023-06-23 at 10 51 31 AM

ios/NativeBridge.swift
```diff
@@ -0,0 +1,220 @@
...
+            case .idTokenValidationFailed: code = "ID_TOKEN_VALIDATION_FAILED"
+            case .other: code = "OTHER"
+            default: code = "UNKNOWN"
+        }
+        return code
+    }
+}
+
+extension CredentialsManagerError {
+    func getReactNativeErrorCode() -> String {
```
Widcket, line 206: Same as above.

ios/NativeBridge.swift
poovamraj: The Java dev in me got the best of me there 😅

ios/NativeBridge.swift
poovamraj: Fixed

src/webauth/index.ts
```diff
@@ -40,13 +24,11 @@ const callbackUri = (domain: string, customScheme?: string) => {
  * @see https://auth0.com/docs/api-auth/grant/authorization-code-pkce
  */
 class WebAuth {
-  private client;
   private domain;
   private clientId;
   private agent;
```
Widcket, line 29: Since now it's all being handled by Auth0.Android and Auth0.swift, do we still need the `agent`?

src/webauth/__tests__/webauth.spec.js
```diff
@@ -1,158 +1,73 @@
...
-        );
-      const showMock = jest
-        .spyOn(webauth.agent, 'show')
-        .mockImplementation(authorizeUrl => ({
-          then: () => Promise.resolve(authorizeUrl),
-        }));
-      const parameters = {
-        invitationUrl:
-          'https://myapp.com/?invitation=inv123&organization=org123',
+  describe('authorize', () => {
```
Widcket, line 16: Should we add a test case with only the required parameters? We already have a test case with all possible parameters.

src/webauth/__tests__/webauth.spec.js
```diff
@@ -1,158 +1,73 @@
...
+        { ...parameters, ...options }
       );
-      newTransactionMock.mockRestore();
       showMock.mockRestore();
     });
+  });
 
-    it('should build the callback URL with a custom scheme when logging out', async () => {
-      const parameters = {customScheme: 'custom-scheme'};
+  describe('clearSession', () => {
```
Widcket, line 56: Same as above.

src/webauth/__tests__/webauth.spec.js
Widcket: Also, should we add a test case with no parameters, to verify it produces an error?

ios/A0Auth0.m
```diff
@@ -1,31 +1,13 @@
...
-@property (weak, nonatomic) SFSafariViewController *last;
-@property (strong, nonatomic) NSObject *authenticationSession;
-@property (strong, nonatomic) CredentialsManagerBridge *credentialsManagerBridge;
-@property (copy, nonatomic) RCTResponseSenderBlock sessionCallback;
-@property (assign, nonatomic) BOOL closeOnLoad;
+@interface A0Auth0 ()
+@property (strong, nonatomic) NativeBridge *nativeBridge;
 @end
 
 @implementation A0Auth0
```
Widcket, line 13: Is the `methodQueue` used?

Widcket (changes requested):
Almost done, just a comment left. Good work!

src/webauth/index.ts
poovamraj: Looks like this has been always the case, fixed this as a part of the major release and added it to breaking change notes ([commit 1](https://github.com/auth0/react-native-auth0/pull/654/commits/d91096675b9d33efa7a9180d02447f028b1dfd14), [commit 2](https://github.com/auth0/react-native-auth0/pull/654/commits/2ab4104de3edf1646a055485ef9b8974ddfeead9))

Widcket (approved):

Issue #684, opened 2023-07-28
Unable to change the access token time


AbhijeetVelotio:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

we are using offline_access scope but we are unable to make the token expiration to 7 days
Changd it in https://manage.auth0.com/dashboard/us/dev-w57h3bsvynubw8ud/apis/64c2649165a24891edd206f6/settings
Application > API's > Auth0 Management API
<img width="1051" alt="Screenshot 2023-07-28 at 7 50 32 PM" src="https://github.com/auth0/react-native-auth0/assets/140828627/433278c5-0da0-41bc-9eb7-768fe77cbab7">

But in Auth response with passwordRealm 

const user_auth = await auth0.auth.passwordRealm({
                username: userCredentials?.email,
                password: userCredentials?.password,
                realm: 'Auth',
                scope: 'openid profile email offline_access'
})

<img width="978" alt="Screenshot 2023-07-28 at 7 53 50 PM" src="https://github.com/auth0/react-native-auth0/assets/140828627/296f09ff-ab7b-4da0-979d-8e27cee3915e">

expiresIn time is not changing 

Can anyone help 
Thank in advance!

            
            
            

### Reproduction

Step 1 : Go to Application > API's > Auth0 Management API change the Access token Expiration time
Step 2: Login with auth0.auth.passwordRealm method

### react-native-auth0 version

2.17.4

### React Native version

0.68.2

### Platform

Android

### Platform version(s)

10

poovamraj:
[...]

PR #688, opened 2023-08-09
Rename expires in to expires at

poovamraj:
### Changes
We are changing all use of `expiresIn` to `expiresAt` as a UNIX timestamp to make it consistent. These changes are done in the API client as well. This will avoid confusions and use proper naming convention and we will use the major release as an opportunity to make this change.

### References
https://github.com/auth0/react-native-auth0/issues/684

### Testing
Manual testing done in Android and iOS as well

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -12,7 +12,7 @@ export type Credentials = {
    */
   tokenType: string;
   /**
-   * Used to denote when the token will expire as UNIX timestamp
+   * Used to denote when the token will expire, as a UNIX timestamp
    */
   expiresAt: number;
   /**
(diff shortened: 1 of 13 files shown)


src/types.ts
```diff
@@ -12,9 +12,9 @@ export type Credentials = {
    */
   tokenType: string;
   /**
-   * Used to denote when the token will expire from the issued time
+   * Used to denote when the token will expire as UNIX timestamp
```
Widcket, line 15:
```suggestion
   * Used to denote when the token will expire, as a UNIX timestamp
```

Widcket (dismissed):

Widcket (approved):

Issue #700, opened 2023-08-15
Incorrect parameters to `authorize` in docs


objectiveSee:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

In README, the example you have provided for using `authorize()` is not correct. There are two required parameters which cannot be omited in Typescript.

https://github.com/auth0/react-native-auth0/blob/master/README.md?plain=1#L280

The correct usage would be:
```
authorize({},{})
```

Empty objects must be passed because `parameters` and `options` are not optional. 

Solution: Update docs. Or, update `authorize` to take a single object with optional parameters such as:

Change this:
```
  authorize: (
    parameters: WebAuthorizeParameters,
    options: WebAuthorizeOptions
  ) => Promise<Credentials | undefined>;
```

to this:
```
  authorize: ({
        parameters?: WebAuthorizeParameters,
        options?: WebAuthorizeOptions
    }
  ) => Promise<Credentials | undefined>;
```

### Reproduction

See README

### react-native-auth0 version

3.0.0

### React Native version

72.3

### Expo version

49

### Platform

Android, iOS

### Platform version(s)

any

objectiveSee:
`clearSession` has the same bug in docs: https://github.com/auth0/react-native-auth0#expo

poovamraj:
@objectiveSee thanks a lot for raising this. We will make the API cleaner to use.

poovamraj:
@felipem292 can you create a seperate issue with more details?

poovamraj:
[...]

PR #701, opened 2023-08-15
Make `authorize` and `clearSession` parameters optional

poovamraj:
### Changes
The types are having mandatory parameters for `authorize` and `clearSession`. We have fixed this by making them optional as it should be.

### References
https://github.com/auth0/react-native-auth0/issues/700

### Testing


objectiveSee:
Thanks for the quick response @poovamraj !! 👏👏

Issue #785, opened 2023-10-30
`webAuth.authorize` promise not completing on an Android deep link


rick-lannan-upgrowth:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

In v3 versions of the package, there is an Android bug where the `webAuth.authorize` promise never completes when the auth0 overlay gets hidden when the app is minimised and reopened via a deep link or the app icon. The result of this is code awaiting the auth0 authorize action just sits waiting forever

I have managed to patch the issue and would like the patch applied back to the source code in the next release. Here is the patch to `src/webauth/agent.ts` that resolves the issue:

```
import { AppState, AppStateStatus, NativeEventSubscription, NativeModules, Platform } from 'react-native';
import { Credentials } from 'src/types';
import { _ensureNativeModuleIsInitialized } from '../utils/nativeHelper';
import {
  AgentLoginOptions,
  AgentLogoutOptions,
  AgentParameters,
  Auth0Module,
} from 'src/internal-types';

type AuthError = {
  code: string
  message: string
}
const A0Auth0: Auth0Module = NativeModules.A0Auth0;
class Agent {
  async login(
    parameters: AgentParameters,
    options: AgentLoginOptions
  ): Promise<Credentials> {
    if (!NativeModules.A0Auth0) {
      return Promise.reject(
        new Error(
          'Missing NativeModule. React Native versions 0.60 and up perform auto-linking. Please see https://github.com/react-native-community/cli/blob/master/docs/autolinking.md.'
        )
      );
    }
[...]

PR #798, opened 2023-11-08
Handle incomplete promise in web authentication

poovamraj:
### Changes
We have a issue in our RNA where the promise for web authentication remains incomplete if the user moves out of the app in between the login.

This is happening because React Native opens the root instance of a [task](https://developer.android.com/guide/components/activities/tasks-and-back-stack) because it uses launch mode as [singleTask](https://developer.android.com/guide/components/activities/tasks-and-back-stack) which usually doesn't happen in Android apps.

To fix this we handle the `onNewIntent` and reject the promise that the user has cancelled the authentication.

### References
https://github.com/auth0/react-native-auth0/issues/785

### Testing
The changes have been tested manually


Widcket (dismissed):

rick-lannan-upgrowth:
Hi @poovamraj, if I understand correctly, this PR forces a cancellation of the promise when the auth window is closed prematurely. Whilst this is valid for unknown events which cause the auth window to be closed, I would argue that stopping the auth window from being closed when the app is backgrounded and relaunched via the app icon or a deep link is also required (you mentioned on https://github.com/auth0/react-native-auth0/issues/785 this should be possible). The reasoning is it's completely normal for a user to background the app during an auth flow to open their email or sms app if there is an auth0 action that requires email verification or 2FA enrolment as part of the login flow

For example, if email verification is required on registration, if the user does the verification on the same device and then relaunches the app via the app icon, this would result in a broken registration process where the auth window disappears and the app is in an a signed out state. And it would not be evident to users that their account has actually been created already and they need to do a Sign In and not another Register. I know this to be the case as this is what happens in the v2 version of the package as it does something similar with the `a0.session.user_cancelled` error

In our current implementation using the v2 version of the package, we ended up fixing the broken registration flow with a complicated solution looking for the `a0.session.user_cancelled` error and if on Android, attempting a silent auth using `webAuth.authorize` with `{ "prompt": "none, "screen_hint": "login" }` - this results in an automatic recovery of the broken registration flow by automatically signing the user in if they've completed their email verification... but it shouldn't be the responsibility of the consuming application to handle this. Whilst we could just leave that workaround in place it really should be fixed at the package level. This would also make the experience on IOS and Android experience the same

I'd also argue that `a0.session.user_cancelled` shouldn't be the error code as it's not a user triggered cancellation, but rather that the auth overlay was dismissed by the OS. Having a separate code means the calling application can differentiate between a cancel triggered by the user or a cancel due to a failed auth process

poovamraj:
@rick-lannan-upgrowth I understand this and that is the intention of the SDK as well. But there is a limitation in react-native where it forces to create a new task from the root when the app is opened as you can see [here](https://github.com/facebook/react-native/blob/main/packages/react-native/template/android/app/src/main/AndroidManifest.xml#L16). This means no matter what we do in our SDK a new task will be created from the root because of how React Native is configured.

You can override this behaviour by setting the `AndroidManifest.xml` to the following

```
  <application>
      <activity
          android:name=".MainActivity"
          android:label="@string/app_name"
          android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|screenSize|smallestScreenSize|uiMode"
          android:windowSoftInputMode="adjustResize"
          android:exported="true">
          <intent-filter>
              <action android:name="android.intent.action.MAIN" />
              <category android:name="android.intent.category.LAUNCHER" />
          </intent-filter>
      </activity>
  </application>
```

All this basically does is removes the `launchMode` from how React Native sets it and you can achieve the behaviour you are asking for. But this is changing the internal behaviour of React Native and that wouldn't be recommeneded behaviour

About the error message, I agree that this can be a more unique error message. We will update it from our end 👍

Widcket (approved):

Issue #911, opened 2024-05-31
Can't configure library for multi-tenant


jnjdev42:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I followed auth0 documentation and it seems we can't configure the library for multi tenant with mulmti domain and multi redirect uri.

### Reproduction

1. Configure auth0 with adding this line in build.gradle :  
2. manifestPlaceholders = [auth0Domain: "domain", auth0Scheme: "${applicationId}.auth0"]

### Additional context

I tried to remove manifestPlaceholders to add multiple data intent filter into manifest. But, application not compiling. Because it needs manifestPlaceholders "auth0domain" to compile.

### react-native-auth0 version

v3.2.1

### React Native version

0.74.1

### Platform

Android

### Platform version(s)

14

brth31:
@jnjdev42 this is a potential duplicate of https://github.com/auth0/react-native-auth0/issues/747

jnjdev42:
> @jnjdev42 this is a potential duplicate of #747

No. The problem for me is about redirections after login and after logout. Moreover, the problem is to set "static" value inside gradle for a domain. Which is not what I am looking for, as I work with different domains.

poovamraj:
Hey @jnjdev42, you want to use multiple domains for redirection if I understand this issue, have you considered a setup like this - https://github.com/auth0/react-native-auth0/issues/747#issuecomment-2439196254

Or is your callback URL auto-generated so that you can't hardcode them?

Issue #747, opened 2023-10-02
Unable to create a new Auth0Provider with a different domain


chrisbruford:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Create a new `Auth0Provider` with a different `domain` prop still results in directing users to the initial `domain` when using the authorize method provided by `useAuth0` hook.

Worked on 2.17.4

### Reproduction

1. Initialise the `Auth0Provider` with any valid `domain` prop
2. Initialise a new `Auth0Provider` with a new valid `domain` prop
3. Observe that the login function provided by `useAuth0` still directs the browser to the initial domain

### Additional context

We use multiple Auth0 Tenants to isolate users from different customers and in our React Native app we allow the user to select which customer they are from and thus we dynamically change the `domain` being sent to the Auth0 SDK. 

This worked well up until we upgraded to v3 and now it fails to change the domain. I have traced it through the auth0 SDK and found the domain change is successfully being propogated all the way to the NativeModule import; but the SDK still directs the browser window to the previously set domain. Seems like the domain gets cached somewhere (perhaps the useAuth0 hook using the old object?)

[...]

PR #931, opened 2024-07-19
feat: added support for domain switching

desusai7:
### Added support for domain switching

* Updated the `hasValidAuth0Instance` method in both `Android` & `iOS` modules to check if the existing auth0 instance `clientId` & `domain` are same as the `clientId` & `domain` passed from the React Native layer and if they are different we are creating new instance of `Auth0`, this would enable customers to switch between different domains quickly. 

* To support multiple domains in an Android app, developers need to add intent filters to their `AndroidManifest.xml` file. The following code snippet demonstrates how to do this. By incorporating these changes, the default `IntentFilter` provided by the library is overridden, eliminating the need to specify `manifestPlaceHolders` in the app's build.gradle file. Simply add a `<data>` element within the `<intent-filter>` of `com.auth0.android.provider.RedirectActivity` for each domain you wish to accommodate.

```xml
    <activity
      android:name="com.auth0.android.provider.RedirectActivity" tools:node="replace"
      android:exported="true">
      <intent-filter>
        <action android:name="android.intent.action.VIEW" />

        <category android:name="android.intent.category.DEFAULT" />
        <category android:name="android.intent.category.BROWSABLE" />

        <data
          android:host="domain1"
          android:pathPrefix="/android/${applicationId}/callback"
          android:scheme="${applicationId}.auth0" />
        <data
          android:host="domain2"
          android:pathPrefix="/android/${applicationId}/callback"
          android:scheme="${applicationId}.auth0" />
      </intent-filter>
    </activity>
```

### References

#747 
#911 

### Testing

Since these changes are made on the bridge layers of both Android and iOS platforms, we do not have test setup there, so updated the sample app to use multiple domains in two different screens by default and tested it on both Android and iOS platforms.

### Checklist

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.java
@@ -63,13 +63,13 @@ public class A0Auth0Module extends ReactContextBaseJavaModule implements Activit
     }
 
     @ReactMethod
-    public void hasValidAuth0Instance(Promise promise) {
-        promise.resolve(this.auth0 != null && this.secureCredentialsManager != null);
+    public void hasValidAuth0Instance(String clientId, String domain, Promise promise) {
+        promise.resolve(this.auth0 != null && this.auth0.getClientId().equals(clientId) && this.auth0.getDomainUrl().contains(domain));
     }
 
     @ReactMethod
     public void getCredentials(String scope, double minTtl, ReadableMap parameters, boolean forceRefresh, Promise promise) {
-        Map<String,String> cleanedParameters = new HashMap<>();
+        Map<String, String> cleanedParameters = new HashMap<>();
         for (Map.Entry<String, Object> entry : parameters.toHashMap().entrySet()) {
             if (entry.getValue() != null) {
                 cleanedParameters.put(entry.getKey(), entry.getValue().toString());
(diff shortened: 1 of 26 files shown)


android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -63,13 +63,13 @@ public void initializeAuth0(String clientId, String domain) {
     }
 
     @ReactMethod
-    public void hasValidAuth0Instance(Promise promise) {
-        promise.resolve(this.auth0 != null && this.secureCredentialsManager != null);
+    public void hasValidAuth0Instance(String clientId, String domain, Promise promise) {
+        promise.resolve(this.auth0 != null && this.auth0.getClientId().equals(clientId) && this.auth0.getDomainUrl().contains(domain));
```
poovamraj, line 67: `contains` check needs to be reworked

poovamraj (dismissed):

poovamraj (dismissed):

poovamraj (approved):

MAlvesKitestring:
Hey, guys! @desusai7  @poovamraj  Great feature, thanks for working on it! 

We are using expo, do you guys know how we could use it? Is it available through some config plugin property?

desusai7:
Hi @MAlvesKitestring, 

We've released this just now as part of version [4.0.0-beta.0](https://www.npmjs.com/package/react-native-auth0/v/4.0.0-beta.0), yes, we are supporting expo as well, please check out more on it [here](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md#expo)

PR #1037, opened 2025-01-24
fix: Interop Layer

guabu:
This PR removes the `codegenConfig` which was caused builds to fail when running under the new architecture. This is also inline with the reactwg recommendation:

> If you have already added a Codegen spec to your library, but the library is not fully converted to TurboModules/Fabric, we recommend that you delete it before proceeding. You can add this back when you are ready to convert your library to natively use TurboModules/Fabric without the Interop Layer.

The changes have been tested with a sample app using the following React Native versions:

* 0.77.0
* 0.76
* 0.70.15

A follow-up PR will contain a complete migration to Turbo Modules.

The release is versioned as `4.1.0-interop.0` and should be tagged as `@interop`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-auth0",
   "title": "React Native Auth0",
-  "version": "4.1.0-interop.0",
+  "version": "4.0.0",
   "description": "React Native toolkit for Auth0 API",
   "main": "lib/commonjs/index",
   "module": "lib/module/index",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-auth0",
   "title": "React Native Auth0",
-  "version": "4.0.0",
+  "version": "4.1.0-interop.0",
   "description": "React Native toolkit for Auth0 API",
   "main": "lib/commonjs/index",
   "module": "lib/module/index",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "react-native-auth0",
   "title": "React Native Auth0",
-  "version": "4.0.0",
```
poovamraj, line 4: We can remove this change as our release pipeline will take care of this for us :) This will remove the lock file changes as well.

poovamraj (changes requested):

package.json
guabu: Done ✅ Thank you!

EddyVinck:
Hello, I don't see this published yet when trying to install

```
yarn add react-native-auth0@4.1.0-interop.0
```

When can we expect it? I wanted to try it already

poovamraj:
@EddyVinck  we are waiting for community feedback but this will be released early next week. You can test this using npm command

`npm install "https://github.com/guabu/react-native-auth0.git#fix-interop" --save`

EddyVinck:
> @EddyVinck we are waiting for community feedback but this will be released early next week. You can test this using npm command
> 
> `npm install "https://github.com/guabu/react-native-auth0.git#fix-interop" --save`

Thanks! I just tested it with a fresh [Ignite](https://github.com/infinitered/ignite) project and it appears to build on my Android emulator using the new architecture.

Manikandan-saminathan:
@poovamraj I tested on both the new and old arch and the build was successful.

Manikandan-saminathan (approved):

poovamraj (approved):

Issue #923, opened 2024-06-17
Platform-agnostic errors and types


thyming:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

I would like to be able to handle common and expected errors in the auth flow gracefully. For example, if the user closes the login popup, this throws an error, or if a refresh token is expired, this throws an error. However, the error codes and structures are undocumented, platform-specific, and untyped in typescript. 

### Describe the ideal solution

The error object returned by `useAuth0` should be a typed error object with enumerated error codes that are not platform-specific except when strictly necessary.  The [react SDK](https://github.com/auth0/auth0-react/blob/182f4ab/src/errors.tsx#L7) does something like this without having the typed error codes, but at least you can reference the oauth error codes from the spec. 

One can imagine at least two error classes that this library could implement:
* OAuth error, exactly like the react one, bonus points for enumerating some of the more common spec'ed codes
* CredentialsManager error, perhaps containing an OAuth error as a cause (e.g. refresh token expiration from the server), or something else related to retrieval

### Alternatives and current workarounds

The current workaround that we use in our codebase is that we have inspected errors that come out in these cases and reverse engineered the error API from that. 
Something like:
```typescript
error && 'code' in error && (error.code === 'a0.session.user_cancelled' || error.code === 'USER_CANCELLED')
```
which feels quite brittle as an undocumented, platform-specific API.

brunezkey:
I faced the same problem. I wanted to anticipate which errors we might encounter from the `authorize` call, [but the error types are non-existent](https://github.com/auth0/react-native-auth0/blob/master/src/hooks/auth0-context.ts#L130C10-L130C15).

[...]

PR #1076, opened 2025-02-14
Expose generic error codes for CredentialsManagerError

kailash-b:
#### Added
We now expose platform agnostic generic error codes that represent different exception scenarios common in both android and iOS. 
We have considered only the `CredentialsManagerError` for now. 
Keeping the backward compatibility in mind, a new field with the generic error code has been exposed via the `BaseError` class.

The users can leverage the field as below : 

```js
try {
  const credentials = await auth0.credentialsManager.getCredentials();
} catch (error) {
  if(error.generic_error_code == "NO_CREDENTIALS"){
    // handle scenario
  }
}
```
**Note** : The old behaviour still remains untouched. 

### References

- The changes partially address the ask in #923 
- [SDK-5643](https://auth0team.atlassian.net/browse/SDK-5643)

### Testing

### Checklist

[SDK-5643]: https://auth0team.atlassian.net/browse/SDK-5643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.java
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.java
@@ -30,6 +30,43 @@ import java.util.Map;
 
 public class A0Auth0Module extends ReactContextBaseJavaModule implements ActivityEventListener {
 
+    public static final Map<CredentialsManagerException, String> ERROR_CODE_MAP = new HashMap<>() {{
+        put(CredentialsManagerException.Companion.getINVALID_CREDENTIALS(), "INVALID_CREDENTIALS");
+        put(CredentialsManagerException.Companion.getNO_CREDENTIALS(), "NO_CREDENTIALS");
+        put(CredentialsManagerException.Companion.getNO_REFRESH_TOKEN(), "NO_REFRESH_TOKEN");
+        put(CredentialsManagerException.Companion.getRENEW_FAILED(), "RENEW_FAILED");
+        put(CredentialsManagerException.Companion.getSTORE_FAILED(), "STORE_FAILED");
+        put(CredentialsManagerException.Companion.getREVOKE_FAILED(), "REVOKE_FAILED");
+        put(CredentialsManagerException.Companion.getLARGE_MIN_TTL(), "LARGE_MIN_TTL");
+        put(CredentialsManagerException.Companion.getINCOMPATIBLE_DEVICE(), "INCOMPATIBLE_DEVICE");
+        put(CredentialsManagerException.Companion.getCRYPTO_EXCEPTION(), "CRYPTO_EXCEPTION");
+        put(CredentialsManagerException.Companion.getBIOMETRIC_ERROR_NO_ACTIVITY(), "BIOMETRIC_NO_ACTIVITY");
+        put(CredentialsManagerException.Companion.getBIOMETRIC_ERROR_STATUS_UNKNOWN(), "BIOMETRIC_ERROR_STATUS_UNKNOWN");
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -30,6 +30,43 @@
 
 public class A0Auth0Module extends ReactContextBaseJavaModule implements ActivityEventListener {
 
+    public static final Map<CredentialsManagerException, String> ERROR_CODE_MAP = new HashMap<>() {{
```
pmathew92, line 33: This can be a private field as we are not exposing this outside of this class and may not be a static one

src/utils/baseError.ts
```diff
@@ -1,12 +1,14 @@
 class BaseError extends Error {
-  constructor(name: string, message: string) {
+  generic_error_code: string;
```
poovamraj, line 2: Let's use field name like
- `identifier`
- `reason`
- `type`

Or something that will be generic and can fit in a single word. 

The purpose of the field will be extensively documented so it would be a one time learning that it is a platform agnostic error code.

src/hooks/reducer.ts
```diff
@@ -1,11 +1,12 @@
+import BaseError from 'src/utils/baseError';
 import { User } from '../types';
 import { deepEqual } from '../utils/deepEqual';
 import { AuthState } from './auth0-context';
 
 type Action =
   | { type: 'LOGIN_COMPLETE'; user: User }
   | { type: 'LOGOUT_COMPLETE' }
-  | { type: 'ERROR'; error: Error }
+  | { type: 'ERROR'; error: BaseError }
```
poovamraj, line 9: What value do we provide to `generic_error_code` if for authentication errors.

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -30,6 +30,43 @@
 
 public class A0Auth0Module extends ReactContextBaseJavaModule implements ActivityEventListener {
 
+    private final Map<CredentialsManagerException, String> ERROR_CODE_MAP = new HashMap<>() {{
+        put(CredentialsManagerException.Companion.getINVALID_CREDENTIALS(), "INVALID_CREDENTIALS");
```
poovamraj, line 34: There is `Enum.name` field provided by Enums in Kotlin, can we use that to simplify the code? I haven't fired up my IDE to test this but looks like there is an opportunity to clean this one.

poovamraj (changes requested):

src/hooks/reducer.ts
kailash-b: It is left empty. 
The field is currently populated only in case of a CredentialManagerException. 
We can re-use this field when we collate the Authentication Errors as well.

android/src/main/java/com/auth0/react/A0Auth0Module.java
kailash-b: The Enums are internal in the underlying android SDK right now and we cannot access them. 
We are leveraging the Public fields exposed via the Companion class.
However, we have a plan to migrate this .java code to .kt. Once that is done, we can leverage kotlin semantics and clean this snippet (tracked as part of [SDK-5699](https://auth0team.atlassian.net/browse/SDK-5699))

poovamraj (approved):

PR #1083, opened 2025-02-17
Added FAQ for mixed case application id

pmathew92:


FAQ.md
```diff
@@ -257,3 +258,7 @@ You can invoke `cancelWebAuth()` to manually clear the current login transaction
 #### Avoid the login/logout alert box
 
 If you don't need SSO, consider using `ephemeral sessions` or `SFSafariViewController` instead of `ASWebAuthenticationSession`. See [2. How can I disable the iOS _login_ alert box?](#2-how-can-i-disable-the-ios-login-alert-box) for more information.
+
+## 9. How can I prevent the autogenerated redirect_uri from breaking if the applicationId has mixed cases in it on Android ?
+
+It is recommended to have your applicationId in lower case to prevent any mismatch with the generated redirect_uri. But in the scenario where you require your applicationId to be of mixed case, to avoid any mismatch , the user can pass a redirectUri whihc matches the one provided in the manage dashboard as part of the `AgentLoginOptions` property.
```
poovamraj, line 264:
```suggestion
It is recommended to have your applicationId in lower case to prevent any mismatch with the generated redirect_uri. But in the scenario where you require your applicationId to be of mixed case, to avoid any mismatch , the user can pass a redirectUri whihc matches the one provided in the manage dashboard as part of the `AgentLoginOptions` property.
```
It is recommended to have your applicationId in lower case without special characters to prevent any mismatch with the generated redirect_uri. But in the scenario where you require your applicationId to be of mixed case, to avoid any mismatch, the user can pass a `redirectUri` which matches the one provided in the manage dashboard as part of the `AgentLoginOptions` property.

FAQ.md
```diff
@@ -257,3 +258,7 @@ You can invoke `cancelWebAuth()` to manually clear the current login transaction
 #### Avoid the login/logout alert box
 
 If you don't need SSO, consider using `ephemeral sessions` or `SFSafariViewController` instead of `ASWebAuthenticationSession`. See [2. How can I disable the iOS _login_ alert box?](#2-how-can-i-disable-the-ios-login-alert-box) for more information.
+
+## 9. How can I prevent the autogenerated redirect_uri from breaking if the applicationId has mixed cases in it on Android ?
```
poovamraj, line 262:
```suggestion
## 9. How can I prevent the autogenerated redirect_uri from breaking if the applicationId has mixed cases in it on Android ?
```
## 9. How can I prevent the autogenerated redirect_uri from breaking if the applicationId has mixed cases or special characters in it on Android ?

poovamraj (changes requested):

FAQ.md
poovamraj, line 264:
```suggestion
It is recommended to have your applicationId in lower case without special characters to prevent any mismatch with the generated redirect_uri. But in the scenario where you require your applicationId to be of mixed case, to avoid any mismatch, the user can pass a `redirectUri` which matches the one provided in the manage dashboard as part of the `AgentLoginOptions` property.
```

poovamraj (changes requested):

FAQ.md
pmathew92: Done

poovamraj (approved):

Issue #1081, opened 2025-02-14
Invalid Expo config plugin in v4.3.0


KimmoHernborg:
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

When running `expo install react-native-auth0` or `expo prebuild` I get the following error message:

```
CommandError: Package "react-native-auth0" does not contain a valid config plugin.
Learn more: https://docs.expo.dev/guides/config-plugins/#creating-a-plugin

Unexpected token 'typeof'
```
Cannot create a working build with version 4.3.0, but earlier versions work.

### Reproduction

1. Install version v4.3.0 (latest at the time of writing) into project `expo install react-native-auth0`
```
CommandError: Package "react-native-auth0" does not contain a valid config plugin.
Learn more: https://docs.expo.dev/guides/config-plugins/#creating-a-plugin

Unexpected token 'typeof'
```
2. Prebuild to update native folders `expo prebuild`
```
CommandError: Package "react-native-auth0" does not contain a valid config plugin.
Learn more: https://docs.expo.dev/guides/config-plugins/#creating-a-plugin

Unexpected token 'typeof'
```
3. Trying to run either a `expo run:ios` or `expo run:android` fails with the following error:
```
The following build commands failed:
[...]

Issue #804, opened 2023-11-20
newArchEnabled: Cannot specify link libraries for target "react_codegen_RNAuth0Spec" which is not built by this project.


elroyshilling:
### Checklist

- [X] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I have enabled **newArchEnabled** and builds are failing for Android. 

**Debugging**: I can build the app successfully by removing react-native-auth0.
**Version**: 3.0.2
**Issue**: react_codegen_RNAuth0Spec
**Logs**:

CMake Error /node_modules/react-native/ReactAndroid/cmake-utils/ReactNative-application.cmake:124 (target_link_libraries):
    Cannot specify link libraries for target "react_codegen_RNAuth0Spec" which
    is not built by this project.
  Call Stack (most recent call first):

---------------------------

> Task :app:configureCMakeDebug[arm64-v8a] FAILED
w: Detected multiple Kotlin daemon sessions at build/kotlin/sessions

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

---------------------------

### Reproduction

1.  Enable newArchEnabled

> Build fails

3. Disable newArchEnabled

> Build successfully 

4. Remove react-native-auth0 & Enable  newArchEnabled

> Build successfully 

### react-native-auth0 version

3.0.2

### React Native version

0.72.7

### Platform

Android

### Platform version(s)

0.72.7

cj-hnadeem:
[...]

PR #1097, opened 2025-03-10
Dependency and config update as per react-native-builder-bob

subhankarmaiti:
### Changes
- Added ESM support to react-native-builder-bob
- Dependency updated as per react native `0.77.1`

### References
- #1081
- #804 
- [ESM support](https://callstack.github.io/react-native-builder-bob/esm)

### Checklist


src/credentials-manager/index.ts
```diff
@@ -51,8 +51,8 @@ class CredentialsManager {
     } catch (e) {
       const json = {
         error: 'a0.credential_manager.invalid',
-        error_description: e.message,
-        code: e.code,
+        error_description: (e as NativeModuleError).message,
+        code: (e as NativeModuleError).code,
```
pmathew92, line 55: I hope this is the new way to access the error message post the version updates and it will continue working as it was and won't break anything

src/hooks/auth0-provider.tsx
```diff
@@ -100,7 +96,7 @@ const Auth0Provider = ({
             user = getIdTokenProfileClaims(credentials.idToken);
           }
         } catch (error) {
-          dispatch({ type: 'ERROR', error });
+          dispatch({ type: 'ERROR', error: error as BaseError });
```
pmathew92, line 99: Why is this casting required ?

pmathew92:
Hi @subhankarmaiti  Does updating the expo version fix the snyk vulnerability > If so , is it feasible for us to update the version?

src/credentials-manager/index.ts
subhankarmaiti: Yes. This is just to inform typescript what type of error it is, and should not break anything

src/hooks/auth0-provider.tsx
subhankarmaiti: @pmathew92 its required due to the changes in tsconfig file `"strict": true`. we have to explicitly define all the types

PR #1101, opened 2025-03-13
feat: Support Turbo Modules

guabu:
### Changes

This PR includes a number of changes to convert the existing Legacy Native Modules implementation to Turbo Modules to support the new architecture.

At a high-level, the changes are:

1. Generated the turbo module specs (live under `src/specs/NativeA0Auth0.ts`)
2. Added necessary `codeGen` to the `package.json` scripts
3. Renamed `A0Auth0.m` to `A0Auth0.mm` to support C++ code
4. Updated the podspec file to install the necessary dependencies when using the new arch
5. Updated `build.gradle` to support new arch builds in Android
6. Updated to use the new native specs

⚠️ The changes in this PR are considered as breaking changes to be released as part of a new major version. The minimum supported React Native version has been bumped to `0.76.0` and only supports the new architecture.

### Testing

### Checklist


mazenchami:
@guabu thanks for putting this together. after review the docs on (turbo modules)[https://reactnative.dev/docs/next/turbo-native-modules-introduction]

we can see that steps 1-3 are completed:
step 1: added the specs with `NativeA0Auth0.ts`
step 2: configured the `package.json`
step 3: not required by a library as it is the App code to use the library

But step 4: implementing the native side of the code, was not completed:

For android, `A0OAuthModule` would have to implement the `NativeA0OAuthSpec`. You'd need to also modify the `A0OAuthPackage`.
For iOS, the `A0Auth0` interface needs to conform to `NativeA0OAuthSpec` or inherit from `NativeA0OAuthSpecBase` if they need to emit events, as explained [here](https://reactnative.dev/docs/next/the-new-architecture/native-modules-custom-events).
For iOS, the `package.json` should be updated as explained [here](https://reactnative.dev/docs/next/turbo-native-modules-introduction?platforms=ios#register-the-native-module-in-your-app).

step 4 would be needed for full turbo modules compatibility, to get the full benefits of the new architecture and so the library will work even when support for interop is removed in the future!

subhankarmaiti:
Hi @mazenchami, thank you for the feedback. We're looking into it and will share updates in this thread.

PR #1202, opened 2025-06-09
Enforce Yarn-Only Policy & Auto-Cleanup Forbidden Lock Files

subhankarmaiti:
This PR enforces the use of **Yarn** as the only package manager for this project and introduces automated cleanup for forbidden lock files and node_modules if npm or pnpm is used.

### ✨ What’s Changed

- ➕ **Added** `scripts/ensure-yarn.js`:  
  Checks the package manager during install. If npm or pnpm is detected, it runs `scripts/clean-forbidden.sh` to remove `package-lock.json`, `pnpm-lock.yaml`, and `node_modules` from both the root and `example/` directories, then exits with an error. Also it safely deletes forbidden lock files and node_modules, ignoring errors if files do not exist.
- 🔄 **Updated** the `preinstall` script in `package.json` to invoke `ensure-yarn.js`, ensuring this check runs automatically on every install.

### 🛡️ Why

- Prevents accidental use of npm or pnpm.
- Keeps the workspace clean and consistent with a Yarn-only policy.
- Automatically cleans up unwanted lock files and node_modules if another package manager is used.

### 🧑‍💻 Impact

- Contributors and CI will be blocked from using npm or pnpm.
- Any forbidden files will be automatically cleaned up if attempted.

---

🧹 **No more rogue lock files! Only Yarn is allowed.**

### DIFF
--- a/A0Auth0.podspec
+++ b/A0Auth0.podspec
@@ -11,6 +11,7 @@ Pod::Spec.new do |s|
   s.license      = package['license']
   s.authors      = package['author']
   s.platforms    = { :ios => min_ios_version_supported }
+  s.swift_version = '5.0'
   s.source       = { :git => 'https://github.com/auth0/react-native-auth0.git', :tag => "v#{s.version}" }
 
   s.source_files = 'ios/**/*.{h,m,mm,swift}'
--- a/README.md
+++ b/README.md
@@ -664,6 +664,32 @@ _Note_ : We have platform agnostic error codes available only for `CredentialsMa
 | `NO_NETWORK`          | `NO_NETWORK`                                                                                                                                                                                                                                                                                                                                     |                                 |
 | `API_ERROR`           | `API_ERROR`                                                                                                                                                                                                                                                                                                                                      |                                 |
 
+## Troubleshooting
+
+### Swift 6 Compatibility Issues on iOS
+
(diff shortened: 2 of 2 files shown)


scripts/clean-forbidden.sh
gyaneshgouraw, line 1: We should double check, this will accurately work in cross platform environment like windows.
Will `rm` commands behave exactly in the same manner on windows like they work on mac.

scripts/ensure-yarn.js
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
+// scripts/ensure-yarn.js
+if (!process.env.npm_config_user_agent?.includes('yarn')) {
+  require('child_process').execSync('sh scripts/clean-forbidden.sh', {
+    stdio: 'inherit',
+  });
+  console.error(
+    'Error: Only yarn is allowed. Please use yarn to install dependencies.'
```
gyaneshgouraw: How about a more contextual message with exact command in error mesage

> This project uses Yarn. Please use `yarn install` instead of npm or pnpm to avoid dependency issues.

And also adding a warning message.
> We have removed `node_modules` and lockfiles from other package managers like npm

gyaneshgouraw-okta (dismissed):

scripts/clean-forbidden.sh
subhankarmaiti: instead of rm used node for deleting files

scripts/ensure-yarn.js
subhankarmaiti: Updated the message with more contextual data

gyaneshgouraw-okta (approved):

PR #1204, opened 2025-06-10
Fix: Turbo Module & New Architecture Support!

subhankarmaiti:
This PR migrates our native modules to **Turbo Modules**, enabling support for the React Native **New Architecture**! 🏎️💨

**Key Changes:**

*   **iOS & Android:**
    *   Implemented Turbo Module interfaces (`NativeA0Auth0Spec`).
    *   Updated build configurations (`.podspec`, `build.gradle`) for New and Old Arch compatibility.
    *   Refactored native code (Objective-C/Swift & Java) to align with Codegen specs.
*   **✨ Benefits:**
    *   Improved performance & faster startup.
    *   Enhanced type safety between JS & Native.
    *   Future-proofs the SDK for upcoming RN versions.

**References**

- https://reactnative.dev/docs/turbo-native-modules-introduction
- https://github.com/reactwg/react-native-new-architecture/blob/main/docs/backwards-compat-turbo-modules.md

This update brings significant improvements to the SDK's foundation! 🛠️

### DIFF
--- a/ios/A0Auth0.mm
+++ b/ios/A0Auth0.mm
@@ -40,49 +40,83 @@ RCT_EXPORT_METHOD(getBundleIdentifier:(nonnull RCTPromiseResolveBlock)resolve re
     resolve([[NSBundle mainBundle] bundleIdentifier]);
 }
 
-#ifdef RCT_NEW_ARCH_ENABLED
-- (void)cancelWebAuth:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject { 
+RCT_EXPORT_METHOD(cancelWebAuth:(RCTPromiseResolveBlock)resolve
+               reject:(RCTPromiseRejectBlock)reject) { 
     [self.nativeBridge cancelWebAuthWithResolve:resolve reject:reject];
 }
 
 
-- (void)clearCredentials:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject { 
+RCT_EXPORT_METHOD(clearCredentials:(RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject) { 
     [self.nativeBridge clearCredentialsWithResolve:resolve reject:reject];
 }
 
 
-- (void)getCredentials:(NSString * _Nullable)scope minTTL:(NSInteger)minTTL parameters:(nonnull NSDictionary *)parameters forceRefresh:(BOOL)forceRefresh resolve:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject { 
+RCT_EXPORT_METHOD(getCredentials:(NSString * _Nullable)scope
+                minTTL:(NSInteger)minTTL
+            parameters:(NSDictionary *)parameters
+          forceRefresh:(BOOL)forceRefresh
+               resolve:(RCTPromiseResolveBlock)resolve
+                reject:(RCTPromiseRejectBlock)reject) { 
     [self.nativeBridge getCredentialsWithScope:scope minTTL:minTTL parameters:parameters forceRefresh:forceRefresh resolve:resolve reject:reject];
 }
 
 
-- (void)hasValidAuth0InstanceWithConfiguration:(nonnull NSString *)clientId domain:(nonnull NSString *)domain resolve:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject { 
+RCT_EXPORT_METHOD(hasValidAuth0InstanceWithConfiguration:(NSString *)clientId
+                                        domain:(NSString *)domain
+                                       resolve:(RCTPromiseResolveBlock)resolve
+                                        reject:(RCTPromiseRejectBlock)reject) { 
     BOOL valid = [self checkHasValidNativeBridgeInstance:clientId domain:domain];
     resolve(@(valid));
 }
 
 
-- (void)hasValidCredentials:(NSInteger)minTTL resolve:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject { 
+RCT_EXPORT_METHOD(hasValidCredentials:(NSInteger)minTTL
+                    resolve:(RCTPromiseResolveBlock)resolve
+                     reject:(RCTPromiseRejectBlock)reject) { 
     [self.nativeBridge hasValidCredentialsWithMinTTL:minTTL resolve:resolve];
 }
 
 
-- (void)initializeAuth0WithConfiguration:(nonnull NSString *)clientId domain:(nonnull NSString *)domain localAuthenticationOptions:(nonnull NSDictionary *)localAuthenticationOptions resolve:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject { 
+RCT_EXPORT_METHOD(initializeAuth0WithConfiguration:(NSString *)clientId
+                                  domain:(NSString *)domain
+              localAuthenticationOptions:(NSDictionary * _Nullable)localAuthenticationOptions
+                                 resolve:(RCTPromiseResolveBlock)resolve
+                                  reject:(RCTPromiseRejectBlock)reject) { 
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -28,17 +40,25 @@ def DEFAULT_TARGET_SDK_VERSION = 34
 android {
   compileSdkVersion safeExtGet('compileSdkVersion', DEFAULT_COMPILE_SDK_VERSION)
   buildToolsVersion safeExtGet('buildToolsVersion', DEFAULT_BUILD_TOOLS_VERSION)
-  namespace 'com.auth0.react'
+  if (project.android.hasProperty("namespace")) {
+        namespace 'com.auth0.react'
```
pmathew92: Why this if check is needed ?

android/src/main/AndroidManifestNew.xml
```diff
@@ -0,0 +1,6 @@
+<!--Ignore rule required due to the 'queries' tag, fix only available for targetSdk=30 -->
+<!--suppress AndroidElementNotAllowed -->
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
pmathew92: Why do we need to create a new manifest for internet permission ? Auth0.Android repo comes with  the above permission defined

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -84,6 +84,70 @@ public A0Auth0Module(ReactApplicationContext reactContext) {
         this.reactContext.addActivityEventListener(this);
     }
 
+    @ReactMethod
+    @Override
+    public void webAuth(String scheme, String redirectUri, @Nullable String state, @Nullable String nonce, @Nullable String audience, @Nullable String scope, @Nullable String connection, @Nullable Double maxAge, @Nullable String organization, @Nullable String invitationUrl, @Nullable Double leeway, @Nullable Boolean ephemeralSession, @Nullable Double safariViewControllerPresentationStyle, @Nullable ReadableMap additionalParameters, Promise promise) {
+        this.webAuthPromise = promise;
+        Map<String, String> cleanedParameters = new HashMap<>();
+        assert additionalParameters != null;
```
pmathew92: additionalParameters is defined as nullable type so the above check will terminate the program execution when someone passes it as null . Is that what we expect the beahviour to be ? Or does it needs to be handled in a better manner ?

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -84,6 +84,70 @@ public A0Auth0Module(ReactApplicationContext reactContext) {
...
+        if (audience != null) {
+            builder.withAudience(audience);
+        }
+        if (scope != null) {
+            builder.withScope(scope);
+        }
+        if (connection != null) {
+            builder.withConnection(connection);
+        }
+        assert maxAge != null;
```
pmathew92: Same as above comment

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -84,6 +84,70 @@ public A0Auth0Module(ReactApplicationContext reactContext) {
...
+        if (maxAge.intValue() != 0) {
+            builder.withMaxAge(maxAge.intValue());
+        }
+        if (organization != null) {
+            builder.withOrganization(organization);
+        }
+        if (invitationUrl != null) {
+            builder.withInvitationUrl(invitationUrl);
+        }
+        assert leeway != null;
```
pmathew92: Same as the previous comment

android/src/main/java/com/auth0/react/A0Auth0Module.java
```diff
@@ -282,7 +284,7 @@ public void webAuthLogout(String scheme, boolean federated, String redirectUri,
         if (redirectUri != null) {
             builder.withReturnToUrl(redirectUri);
         }
-        builder.start(reactContext.getCurrentActivity(),
+        builder.start(Objects.requireNonNull(reactContext.getCurrentActivity()),
```
pmathew92: This will crash the application on nullable type . Should we handle this in a better manner ?

NandanPrabhu:
@subhankarmaiti  Please include more details on what the PR solves. some links/docs to get a preread before reviewing the PR would be helpful

android/build.gradle
subhankarmaiti: The check was originally added for users with Gradle versions below 7.3, but we can remove it now since React Native 0.79 itself doesn't support those older versions.

android/src/main/AndroidManifestNew.xml
subhankarmaiti: removing it as the namespace check we are removing

android/src/main/java/com/auth0/react/A0Auth0Module.java
subhankarmaiti: Yes, it can be null. added a check for the same

android/src/main/java/com/auth0/react/A0Auth0Module.java
subhankarmaiti: added null check

android/src/main/java/com/auth0/react/A0Auth0Module.java
subhankarmaiti: added null check

android/src/main/java/com/auth0/react/A0Auth0Module.java
subhankarmaiti: The React context is being set at the constructor level. So this will always be not null.

pmathew92 (approved):

android/build.gradle
```diff
@@ -14,9 +14,21 @@ buildscript {
     }
 }
 
+def reactNativeArchitectures() {
+  def value = rootProject.getProperties().get("reactNativeArchitectures")
+  return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
+}
+
+def isNewArchitectureEnabled() {
+  return rootProject.hasProperty("newArchEnabled") && rootProject.getProperty("newArchEnabled") == "true"
```
NandanPrabhu, line 23: @subhankarmaiti  to toggle the newArchEnabled settings it has to be done here for android 
right?

ios/NativeBridge.swift
```diff
@@ -206,12 +206,6 @@ public class NativeBridge: NSObject {
         resolve(credentialsManager.clear())
     }
     
-    @objc public func enableLocalAuthentication(title: String?, cancelTitle: String?, fallbackTitle: String?, evaluationPolicy: Int) {
-        let titleValue = title ?? "Please authenticate to continue"
-        let policyValue = self.convert(policyInt: evaluationPolicy)
-        self.credentialsManager.enableBiometrics(withTitle: titleValue, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle, evaluationPolicy: policyValue)
```
NandanPrabhu, line 212: is this method vital for enabling local authentication? do we have any other method to enable Local authentication?

ios/A0Auth0.mm
```diff
@@ -32,54 +32,103 @@ - (dispatch_queue_t)methodQueue
...
+    if (localAuthenticationOptions.deviceCredentialFallback()) {
+        options[@"deviceCredentialFallback"] = @((NSInteger)localAuthenticationOptions.deviceCredentialFallback().value());
+    }
+    [self tryAndInitializeNativeBridge:clientId domain:domain withLocalAuthenticationOptions:options resolver:resolve rejecter:reject];
+}
+
+RCT_EXPORT_METHOD(saveCredentials:(JS::NativeA0Auth0::Credentials &)credentials resolve:(nonnull RCTPromiseResolveBlock)resolve reject:(nonnull RCTPromiseRejectBlock)reject) {
+    NSMutableDictionary *credentialsDict = [NSMutableDictionary dictionary];
+    
+    credentialsDict[@"idToken"] = credentials.idToken();
```
NandanPrabhu: would suggest to have these in string constants

ios/NativeBridge.swift
subhankarmaiti: I have removed it as we were not using this anywhere

ios/A0Auth0.mm
subhankarmaiti: Yes `idToken` is in string format.

NandanPrabhu (approved):

Issue #1232, opened 2025-07-09
Previous Google social login not shown as of 5.0.0-beta.3


CodingItWrong:
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

On iOS, the web view no longer seems to retain any history of the accounts I'm already logged into. In 5.0.0-beta.1 with `additionalParameters: { prompt: "login" }` specified I could sign into Auth0 via a Google account, sign out of Auth0, and then the next time I tried to sign in via Google it would let me choose the Google account I previously signed into. But as of beta.3 I need to sign in with my Google username/password each time.

(Videos attached showing beta 1 and beta 3 behavior)

| Before                                             | After                                             |
| -------------------------------------------------- | ------------------------------------------------- |
| <video src="https://github.com/user-attachments/assets/af3dcffc-e8a8-47a5-8192-24fea4737aea" alt="before_screenshot" width="300" /> | <video src="https://github.com/user-attachments/assets/411a4568-a99e-486c-900b-fdeb1881103e" alt="after_screenshot" width="300" /> |

### Reproduction

1. Clone this branch locally: https://github.com/CodingItWrong/react-native-auth0/tree/beta-3-login-issue
2. Follow instructions to set up Auth0 app config pointing to a config with Google social login and install and run app
3. Sign into a google account
4. Tap "Log Out" to sign back out of Auth0
5. Sign in again

EXPECTED: Google shows a list including the previously-signed-in account and "Use another account"
ACTUAL: Google prompts for username/email address

### Additional context

[...]

PR #1257, opened 2025-07-28
fix(ios): Ensure useSFSafariViewController: true correctly opens SFSafariViewController

subhankarmaiti:
This PR fixes a bug where setting `useSFSafariViewController: true` in the `authorize` options did not open the SFSafariViewController as expected on iOS.

The logic has been updated to correctly apply a default presentation style (`fullScreen`) when the option is passed as a boolean `true`. This aligns the behavior with the documentation and removes the need for workarounds like passing an empty object `{}`.

Fixes: https://github.com/auth0/react-native-auth0/issues/1232

### DIFF
--- a/src/platforms/native/bridge/NativeBridgeManager.ts
+++ b/src/platforms/native/bridge/NativeBridgeManager.ts
@@ -93,7 +93,8 @@ export class NativeBridgeManager implements INativeBridge {
       parameters.invitationUrl,
       options.leeway ?? 0,
       options.ephemeralSession ?? false,
-      presentationStyle ?? 99,
+      presentationStyle ?? 99, // Since we can't pass null to the native layer, and we need a value to represent this parameter is not set, we are using 99.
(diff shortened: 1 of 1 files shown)


src/platforms/native/bridge/NativeBridgeManager.ts
```diff
@@ -97,7 +93,7 @@ export class NativeBridgeManager implements INativeBridge {
       parameters.invitationUrl,
       options.leeway ?? 0,
       options.ephemeralSession ?? false,
-      presentationStyle,
+      presentationStyle ?? 99,
```
NandanPrabhu: @subhankarmaiti  how are we using default value of 99 ?

subhankarmaiti: we are passing 99 as a default value if nothing is passed from the user and on Native layer we are not setting `style` if 99 passed.

NandanPrabhu (approved):

Issue #1260, opened 2025-07-30
v.5.0.0-beta2 and beta4 After login with `authorize` app deep links to the callback url which causes the app to land on the not found screen.


letidev:
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

After upgrading to Expo SDK 53, I upgraded the auth0 package to v5.0.0-beta4 but after taking a look at issue #1259 I downgraded to beta2 as it mentions that it's more stable than beta4. The issue was present with both versions.

Here's the setup:

```
app
|--- (app) <-- protected routes
|    |--- (tabs)
|    |    |--- _layout.tsx
|    |    |--- index.tsx
|    |--- _layout.tsx
|
|--- _layout.tsx
|--- login.tsx
```

In the `RootLayout` which is in `app/_layout.tsx` we have:

```js
export default RootLayout = () => {
  const { isLoggedIn } = useAuthContext()

  console.log("Pathname: ", usePathname())

  return (
   <Stack>
      <Stack.Protected guard={isLoggedIn}>
        <Stack.Screen name="(app)" />
      </Stack.Protected>

      <Stack.Screen name="login" />
    </Stack>
  )
}
```

On the login screen, after `authorize` from `useAuth0` resolves successfully I do `router.replace("/")`

If we monitor the pathname that is console logged in the root layout we observe the following:

```
Pathname: /login
[...]

PR #1261, opened 2025-08-04
feat: unify and improve error handling across platforms

subhankarmaiti:
This PR unifies error handling across Native (iOS/Android) and Web platforms, providing a consistent, platform-agnostic model that addresses the developer experience issues raised in **[#923](https://github.com/auth0/react-native-auth0/issues/923)**.

#### Key Changes:

*   **Platform-Agnostic Errors**: `CredentialsManagerError` and the new `WebAuthError` now map platform-specific and web error codes to a single, consistent set of codes.
*   **New Typed `WebAuthError`**: Introduces a `WebAuthError` class with a `.type` property (e.g., `'USER_CANCELLED'`) for easy, reliable error identification, eliminating the need for `Platform.OS` checks.
*   **Native SDK Upgrades**: Updates `auth0-android` to `3.8.0` and `Auth0.swift` to `2.13.0` to leverage their latest error reporting capabilities.
*   **Updated Documentation**: The `README.md` is updated with new error tables and "Before/After" code examples to guide developers on the new, simplified approach.

Fixes: https://github.com/auth0/react-native-auth0/issues/923, https://github.com/auth0/react-native-auth0/issues/1260

### DIFF
--- a/src/platforms/native/adapters/__tests__/NativeCredentialsManager.errors.spec.ts
+++ b/src/platforms/native/adapters/__tests__/NativeCredentialsManager.errors.spec.ts
@@ -18,67 +18,198 @@ const mockBridge: jest.Mocked<INativeBridge> = {
   resumeWebAuth: jest.fn(),
 };
 
-describe('NativeCredentialsManager Error Handling', () => {
+describe('Common Error Handling - NativeCredentialsManager', () => {
   let manager: NativeCredentialsManager;
 
   beforeEach(() => {
     jest.clearAllMocks();
     manager = new NativeCredentialsManager(mockBridge);
   });
+  describe('Core Error Types', () => {
+    const coreErrorTestCases = [
+      {
+        code: 'INVALID_CREDENTIALS',
+        message: 'Invalid credentials provided.',
+        expectedType: 'INVALID_CREDENTIALS',
+        method: 'getCredentials',
+      },
+      {
+        code: 'NO_CREDENTIALS',
+        message: 'No credentials stored.',
+        expectedType: 'NO_CREDENTIALS',
+        method: 'getCredentials',
+      },
+      {
+        code: 'NO_REFRESH_TOKEN',
+        message: 'No refresh token available.',
+        expectedType: 'NO_REFRESH_TOKEN',
+        method: 'getCredentials',
+      },
+      {
+        code: 'RENEW_FAILED',
+        message: 'Refresh token renewal failed.',
+        expectedType: 'RENEW_FAILED',
+        method: 'getCredentials',
+      },
+      {
+        code: 'STORE_FAILED',
+        message: 'Failed to store credentials.',
+        expectedType: 'STORE_FAILED',
+        method: 'saveCredentials',
+        args: [{}],
+      },
+      {
+        code: 'REVOKE_FAILED',
+        message: 'Failed to revoke credentials.',
+        expectedType: 'REVOKE_FAILED',
+        method: 'clearCredentials',
+      },
+      {
+        code: 'LARGE_MIN_TTL',
+        message:
+          'The minTTL requested is greater than the lifetime of the renewed access token. Request a lower minTTL or increase the "Token Expiration" value in the settings page of your Auth0 API.',
+        expectedType: 'LARGE_MIN_TTL',
+        method: 'getCredentials',
+      },
+      {
+        code: 'CREDENTIAL_MANAGER_ERROR',
+        message: 'General credentials manager error occurred.',
+        expectedType: 'CREDENTIAL_MANAGER_ERROR',
+        method: 'getCredentials',
+      },
+      {
+        code: 'BIOMETRICS_FAILED',
+        message: 'Biometric authentication failed.',
+        expectedType: 'BIOMETRICS_FAILED',
+        method: 'getCredentials',
+      },
+      {
+        code: 'NO_NETWORK',
+        message: 'No network connection available.',
+        expectedType: 'NO_NETWORK',
+        method: 'getCredentials',
+      },
+      {
+        code: 'API_ERROR',
+        message: 'API request failed.',
+        expectedType: 'API_ERROR',
+        method: 'getCredentials',
+      },
+    ];
 
-  it('should convert an Android "NO_CREDENTIALS" error from getCredentials into a CredentialsManagerError', async () => {
-    const nativeError = {
-      code: 'NO_CREDENTIALS',
-      message: 'No credentials stored.',
-    };
-    mockBridge.getCredentials.mockRejectedValue(nativeError);
+    coreErrorTestCases.forEach(
+      ({ code, message, expectedType, method, args = [] }) => {
+        it(`should handle ${code} error`, async () => {
+          const nativeError = { code, message };
+          (
+            mockBridge[method as keyof typeof mockBridge] as jest.Mock
+          ).mockRejectedValue(nativeError);
 
-    await expect(manager.getCredentials()).rejects.toThrow(
-      CredentialsManagerError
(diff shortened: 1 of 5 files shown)


A0Auth0.podspec
```diff
@@ -16,9 +16,7 @@ Pod::Spec.new do |s|
   s.source_files = 'ios/**/*.{h,m,mm,swift}'
   s.requires_arc = true
 
-  s.dependency 'Auth0', '2.10'
-  s.dependency 'JWTDecode', '3.2.0'
-  s.dependency 'SimpleKeychain', '1.2.0'
+  s.dependency 'Auth0', '2.13'
```
pmathew92, line 19: Is JWTDecode and SimpleKeychain dependencies not required anymore ?

subhankarmaiti: The latest version of Auth0.Swift added those libraries as dependencies. So the used project does not have to add those manually.

src/platforms/web/adapters/WebCredentialsManager.ts
```diff
@@ -32,7 +33,7 @@ export class WebCredentialsManager implements ICredentialsManager {
       const claims = await this.client.getIdTokenClaims();
       if (!claims || !claims.exp) {
         throw new AuthError(
-          'IdTokenMissing',
+          'ID_TOKEN_CLAIM_VALIDATION_FAILED',
```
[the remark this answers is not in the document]
subhankarmaiti: Here we are directly using the `ID_TOKEN_CLAIM_VALIDATION_FAILED` as `AuthError`, so no need to use it in the mapping.

README.md
```diff
@@ -649,22 +649,80 @@ try {
...
+| `RENEW_FAILED`        | `RENEW_FAILED`                                                                                                                                                                                                                                                                                                                                   | `renewFailed`      | `missing_refresh_token`             |
+| `STORE_FAILED`        | `STORE_FAILED`                                                                                                                                                                                                                                                                                                                                   | `storeFailed`      | `invalid_grant`, `consent_required` |
```
pmathew92: why is `missing_refresh_token` error from web mapped with RENEW failed from native. Shouldn't his be part of the `NO_REFRESH_TOKEN` mapping ?

pmathew92: Same for the `invalid_grant` `consent_required ` errors . On second thought do we need CredentialsManager Error for Web since CredentialsManager isn't in the Web?

README.md
```diff
@@ -649,22 +649,80 @@ try {
...
+| `FAILED_TO_LOAD_URL`         | The authorization URL could not be loaded in the browser.                                                                                                     | `a0.session.failed_load`        | -                               |                                      |
+| `BROWSER_TERMINATED`         | The browser was closed unexpectedly, likely because the application was relaunched from the home screen while the login was in progress.                      | `a0.session.browser_terminated` | -                               |                                      |
+| `INVALID_STATE`              | The `state` parameter returned from the server did not match the one sent, indicating a potential Cross-Site Request Forgery (CSRF) attack.                   | `access_denied`                 | `OTHER`                         | `state_mismatch`                     |
```
pmathew92, line 716: Was this tested to check if Android would return `access_denied` error on state missing scenario

src/core/models/CredentialsManagerError.ts
```diff
@@ -8,6 +8,18 @@ const ERROR_CODE_MAP: Record<string, string> = {
...
+  CREDENTIAL_MANAGER_ERROR: 'CREDENTIAL_MANAGER_ERROR',
+  BIOMETRICS_FAILED: 'BIOMETRICS_FAILED',
+  NO_NETWORK: 'NO_NETWORK',
+  API_ERROR: 'API_ERROR',
+
+  // --- Web (@auth0/auth0-spa-js) mappings ---
+  login_required: 'NO_CREDENTIALS',
+  consent_required: 'RENEW_FAILED',
+  invalid_grant: 'RENEW_FAILED',
+  missing_refresh_token: 'NO_REFRESH_TOKEN',
```
pmathew92, line 22: Do we need web mappings in CredentialsManager class ?

src/core/models/CredentialsManagerError.ts
```diff
@@ -35,9 +47,6 @@ const ERROR_CODE_MAP: Record<string, string> = {
   BIOMETRIC_ERROR_UNABLE_TO_PROCESS: 'BIOMETRICS_FAILED',
   BIOMETRICS_INVALID_USER: 'BIOMETRICS_FAILED',
   BIOMETRIC_AUTHENTICATION_FAILED: 'BIOMETRICS_FAILED',
```
pmathew92, line 51: Are we sharing detailed stack trace for scenario where we are returning `BIOMETRICS_FAILED` for different scenarios?

src/core/models/WebAuthError.ts
```diff
@@ -0,0 +1,55 @@
...
+  'state_mismatch': 'INVALID_STATE',
+  'login_required': 'ACCESS_DENIED',
+  'timeout': 'TIMEOUT_ERROR',
+
+  // --- Generic Fallbacks ---
+  'a0.invalid_configuration': 'INVALID_CONFIGURATION',
+  'UNKNOWN': 'UNKNOWN_ERROR',
+  'OTHER': 'UNKNOWN_ERROR',
+};
+
```
pmathew92, line 37: Does this mapping cover all scenarios we have in all 3 platforms ?

README.md
subhankarmaiti: On calling `getCredentials()`, it triggers `getTokenSilently()`. If the access token has expired, `getTokenSilently` will then try to use a refresh token. If it discovers at that moment that it doesn't have one, it fails.

subhankarmaiti: You are right, as we do not have any credential manager in web. when we try to get the credential, it calls`getTokenSilently` and if it throws an error `invalid_grant`, which means it fails on the renew. created `consent_required` as a different error.

src/core/models/CredentialsManagerError.ts
subhankarmaiti: Yes. we are trying to refer the corresponding error to a generic mapping.

src/core/models/CredentialsManagerError.ts
subhankarmaiti: Yes, we are returning the actual error along with the other value

src/core/models/WebAuthError.ts
subhankarmaiti: Yes, all the possible errors could occur in the functions we are calling.

README.md
subhankarmaiti: Yes tested in android by changing the state.

README.md
```diff
@@ -649,22 +649,81 @@ try {
...
+| `NO_REFRESH_TOKEN`    | `NO_REFRESH_TOKEN`                                                                                                                                                                                                                                                                                                                               | `noRefreshToken`   |
+| `RENEW_FAILED`        | `RENEW_FAILED`                                                                                                                                                                                                                                                                                                                                   | `renewFailed`      | `missing_refresh_token`, `invalid_grant`,`consent_required` |
```
pmathew92, line 657: `invalid_grant`,`consent_required` Have we thought about how to handle when web returns these two errors when logging in the first time and not renewing ?

pmathew92:
Could we add few more test cases where all 3 platforms send different error types

README.md
subhankarmaiti: @pmathew92 This error won’t occur during login. It comes from the `getTokenSilently` call, which is why we map it as RENEW_FAILED.

pmathew92: @subhankarmaiti  Ok. But I would still suggest to cross check before merging the same

PR #1329, opened 2025-10-01
feat: export AuthError class from main entry point

afonsojramos:
### Changes

As per https://github.com/auth0/react-native-auth0/blob/master/MIGRATION_GUIDE.md#change-2-standardized-autherror-object, we should be exporting AuthError class to allow consumers to perform instanceof checks and better error handling. This aligns with the library's documentation and enables proper error type detection in user code.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,17 +1,21 @@
-import { Auth0ClientFactory } from './factory/Auth0ClientFactory';
 import type { IAuth0Client } from './core/interfaces/IAuth0Client';
+import { Auth0ClientFactory } from './factory/Auth0ClientFactory';
 import type { Auth0Options } from './types';
-export { AuthError } from './core/models/AuthError';
+
+export {
+  AuthError,
+  CredentialsManagerError,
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1,6 +1,7 @@
 import { Auth0ClientFactory } from './factory/Auth0ClientFactory';
 import type { IAuth0Client } from './core/interfaces/IAuth0Client';
 import type { Auth0Options } from './types';
+export { AuthError } from './core/models/AuthError';
```
subhankarmaiti:
```suggestion
export {
  AuthError,
  WebAuthError,
  CredentialsManagerError,
} from '../core/models';
```

We should also export the other error class.

Similar also needs to be added in `src/exports/classes.ts` file for exporting to docs.

afonsojramos: Done! I ran prettier and noticed that it hasn't been run in a while in the repo, as I had a ton of files that were changed. Only committed the ones I touched though.

PR #1412, opened 2025-12-17
feat: add optional scope parameter to clearApiCredentials

subhankarmaiti:
API Changes:
- `clearApiCredentials(audience, scope?)` - Now accepts optional scope parameter

### DIFF
--- a/src/core/interfaces/ICredentialsManager.ts
+++ b/src/core/interfaces/ICredentialsManager.ts
@@ -142,7 +142,7 @@ export interface ICredentialsManager {
    * `getApiCredentials` call for this audience to perform a fresh token exchange.
    *
    * @param audience The identifier of the API for which to clear credentials.
-   * @param scope Optional scope to clear. If not provided, clears all credentials for the audience.
+   * @param scope Optional scope to clear. If credentials were fetched with a scope, it is recommended to pass the same scope when clearing them.
    * @returns A promise that resolves when the credentials are cleared.
    * @throws {CredentialsManagerError} If the operation fails.
    *
@@ -151,7 +151,7 @@ export interface ICredentialsManager {
    * // Clear all credentials for an audience
    * await credentialsManager.clearApiCredentials('https://api.example.com');
    *
-   * // Clear credentials for specific scope
+   * // Clear credentials for specific scope (recommended)
    * await credentialsManager.clearApiCredentials('https://api.example.com', 'read:data');
    * ```
    */
--- a/src/platforms/native/bridge/INativeBridge.ts
+++ b/src/platforms/native/bridge/INativeBridge.ts
@@ -128,7 +128,7 @@ export interface INativeBridge {
    * Optionally filter by scope to clear only specific scope-based credentials.
    *
    * @param audience The audience of the API.
-   * @param scope Optional scope to clear. If not provided, clears all credentials for the audience.
+   * @param scope Optional scope to clear. If credentials were fetched with a scope, it is recommended to pass the same scope when clearing them.
    */
   clearApiCredentials(audience: string, scope?: string): Promise<void>;
 


EXAMPLES.md
```diff
@@ -492,6 +492,14 @@ function MyComponent() {
...
     // Clear cached credentials for a specific API
     await clearApiCredentials('https://first-api.example.com');
+
+    // Or clear with specific scope
+    await clearApiCredentials('https://first-api.example.com', 'read:data');
+  };
+
+  const clearSpecificCredentials = async () => {
+    // You can also use clearCredentials with audience/scope
+    await clearCredentials('https://first-api.example.com', 'read:data');
```
pmathew92: `clearCredentials` is used to clear the default access token credentials. For anything specific with audience/ scope is part of the MRRT flow and should use the `clearApiCredentials` method

EXAMPLES.md
```diff
@@ -531,6 +539,18 @@ console.log('Scope:', apiCredentials.scope);
...
+// Clear with specific scope
+await auth0.credentialsManager.clearApiCredentials(
+  'https://first-api.example.com',
+  'read:data write:data'
+);
+
+// Or use clearCredentials with audience/scope
+await auth0.credentialsManager.clearCredentials(
+  'https://first-api.example.com',
+  'read:data'
```
pmathew92: `clearCredentials` should clear the default access token. Passing the audience and scope to that will cause confusion as we have now two different APIs doing the same work. Ideally clearAPICredentials should be responsible for clearing the token mapped to an audience/scope from the MRRT flow

EXAMPLES.md
subhankarmaiti: Do you suggest not exposing this API for RN users?

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -271,9 +271,14 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
     }
 
     @ReactMethod
-    override fun clearCredentials(promise: Promise) {
-        secureCredentialsManager.clearCredentials()
-        
+    override fun clearCredentials(audience: String?, scope: String?, promise: Promise) {
```
pmathew92: Wouldn't suggest this and keep the default behaviour as it is . Use the `clearApiCredentials` for clearing audience mapped token

android/src/main/oldarch/com/auth0/react/A0Auth0Spec.kt
```diff
@@ -49,7 +49,7 @@ abstract class A0Auth0Spec(context: ReactApplicationContext) : ReactContextBaseJ
 
     @ReactMethod
     @DoNotStrip
-    abstract fun clearCredentials(promise: Promise)
+    abstract fun clearCredentials(audience: String?, scope: String?, promise: Promise)
```
pmathew92: Not recommended

EXAMPLES.md
subhankarmaiti: Another scenario I can think of is when a user calls
`authorize("audience", "read:data")`
and later, in a different flow, calls
`authorize("audience", "write:data")`.

If the user then calls `clearCredentials()`, it would clear only one entry, while the other entry would still remain.

src/platforms/native/adapters/NativeCredentialsManager.ts
```diff
@@ -40,21 +40,12 @@ export class NativeCredentialsManager implements ICredentialsManager {
...
-    await this.handleError(this.bridge.clearCredentials());
-    // Also clear the DPoP key when clearing credentials
-    // Ignore errors from DPoP key clearing - this matches iOS behavior
-    // where we log the error but don't fail the operation
-    try {
-      await this.bridge.clearDPoPKey();
-    } catch {
-      // Silently ignore DPoP key clearing errors
-      // The main credentials are already cleared at this point
-    }
```
subhankarmaiti, line 57: no need to call this. As we are calling it in the native module

Issue #1374, opened 2025-11-10
Users getting logged out with "Unsuccessful Refresh Token exchange, reused refresh token detected"


LA-Johan:
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Our usage of Auth0 is pretty simple. After the user has been logged in we just attach the access token to network calls and verify them server side. 

```
const useApi = (): Api => {
  const { getCredentials } = useAuth0()
  const backend = axios.create({ baseURL: apiUri, timeout: 10000 })
  backend.interceptors.request.use(
    async config => {
      const credentials = await getCredentials()
      config.headers["Authorization"] = `Bearer ${credentials?.accessToken}`
      return config
    },

    async error => {}
  )
}  
```
 
We have gotten user reports that they get error in title. Our understanding is that this error is unrecoverable so we log the user out and force them to re-authenticate. We have a couple of thousand weekly users and this happens to maybe 2% of them. 

One log id as an example:
90020251110035154104236000000000000001223372093331920555

Our top level provider:
```
<Auth0Provider domain={domain} clientId={clientId}>
      <Main />
</Auth0Provider>
```

### Reproduction

Seems to happen randomly, a few times a week for affected users. So far we've only seen it on iOS. 

### Additional context

[...]

PR #1432, opened 2026-01-12
feat: Add maxRetries configuration for credential renewal (iOS)

subhankarmaiti:
Adds automatic retry mechanism with exponential backoff for credential renewal to improve reliability in unstable network conditions. This addresses issue #1374 where refresh token rotation can fail due to transient network errors.

### Changes
- Added `maxRetries` option to `Auth0Options` interface
- Updated iOS native bridge to pass `maxRetries` to `CredentialsManager` initialization
- Updated Android bridge to accept parameter (ignored due to SDK limitation)
- Updated documentation with configuration examples

### Platform Support
- **iOS only**: ✅ Full support via Auth0.swift v2.16.2+

### Usage
```typescript
const auth0 = new Auth0({
  domain: 'YOUR_DOMAIN',
  clientId: 'YOUR_CLIENT_ID',
  maxRetries: 2, // Retry up to 2 times on transient errors (iOS only)
});
```
#### With React Hooks:
```typescript
<Auth0Provider domain="..." clientId="..." maxRetries={2}>
  <App />
</Auth0Provider>
```

#### Automatically retries on:
- Network errors (timeouts, DNS failures)
- Rate limiting (HTTP 429)
- Server errors (HTTP 5xx)

#### Does NOT retry on:
- Invalid/expired/revoked refresh tokens
- Client errors (HTTP 4xx except 429)

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -276,15 +276,15 @@ Automatic retry mechanism for credential renewal to improve reliability in unsta
 
 ### Overview
 
-When your application operates on unstable mobile networks, credential renewal requests may fail due to transient network issues. This is especially problematic with refresh token rotation enabled, as a failed response can invalidate the refresh token even though the server successfully issued new credentials.
-
-The `maxRetries` configuration option enables automatic retry with exponential backoff for the following error scenarios:
+When your application operates on unstable mobile networks, credential renewal requests may fail due to transient network issues. The `maxRetries` configuration option enables automatic retry with exponential backoff for the following error scenarios:
 
 - **Network errors**: Connection timeouts, DNS failures, unreachable hosts
 - **Rate limiting**: HTTP 429 (Too Many Requests)
 - **Server errors**: HTTP 5xx responses
 
-**Example scenario this solves:**
+> **Important:** While the retry mechanism is particularly valuable for refresh token rotation (RRT) scenarios, it can be used to improve credential renewal reliability in any configuration, including non-RRT deployments. The retry logic helps handle transient network failures regardless of your token rotation strategy.
+
+**Example scenario with Refresh Token Rotation:**
 
 1. Request A calls `getCredentials()` and starts a token refresh
 2. Request A successfully hits the server and gets new credentials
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -259,6 +266,188 @@ auth0.auth
...
+5. The retry succeeds within the refresh token rotation overlap window
+
+<a name="credential-renewal-retry-prerequisites"></a>
+
+### Prerequisites
+
+To use the retry mechanism effectively:
+
+1. **SDK Version**: Requires react-native-auth0 v5.4.0 or later
+2. **Refresh Token Rotation**: Should be enabled in your Auth0 tenant
```
pmathew92: Do you want this to be called out ? Agree the major reason to add the support for retry is to handle the scenario for RRT failure but user can use it even for non RRT cases. We can add an important note sections pbly highlighting this and the `Token Overlap Period` point

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -168,8 +168,13 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
         domain: String,
         localAuthenticationOptions: ReadableMap?,
         useDPoP: Boolean?,
+        maxRetries: Double,
```
pmathew92, line 171: `maxRetries` is a property of the CredentialsManager class. I think it would make sense to add this property to the CredentialsManager class configuration than the global Auth0 class similar to how Auth0.Swift is doing along with setting a default value

subhankarmaiti: Agree with the pattern. But we are not maintaining the credential manager class, as the native instance is being created when we initialise the SDK. So the only option to pass the value via `initializeAuth0WithConfiguration`.

pmathew92: Ok.. do you think we can provide an option for the user to provide an instance of CredentialsManager class similar to how we do on `auth0-flutter` https://github.com/auth0/auth0-flutter/blob/35296cbf944811426cd7fd66408a6b18f091ff2f/auth0_flutter/lib/auth0_flutter.dart#L55
This provides us option to move all CredentialManager configuration to this specific class

subhankarmaiti: This requires building a new CredentialManager class and adding support to multiple places. which is not the design pattern we are following. We can do a POC of this approach for the next major version.

EXAMPLES.md
subhankarmaiti: updated the doc

pmathew92 (approved):

PR #1513, opened 2026-04-21
feat(android): expose allowedBrowserPackages option for web authentication

### LINKED ISSUE #392: Firefox users seeing "not found" on Android, both login and logout

mrbrentkelly:
## Changes

Adds an `allowedBrowserPackages` option to `NativeAuthorizeOptions` that restricts which browsers can handle the web authentication flow on Android.

This exposes the browser filtering capability already present in the underlying [Auth0.Android SDK](https://github.com/auth0/Auth0.Android) via `BrowserPicker` and `CustomTabsOptions` through the React Native bridge.

## Motivation

Firefox on Android does not correctly handle App Link redirects, causing the Auth0 login flow to fail silently — the browser renders the callback URL as a page rather than handing off to the app. This is a [known Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1976809).

The same issue was reported and resolved for the Flutter SDK in [auth0/auth0-flutter#392](https://github.com/auth0/auth0-flutter/issues/392) by exposing an `allowedBrowsers` option. This PR brings the same capability to React Native.

## Behaviour

When `allowedBrowserPackages` is set on Android:
- The user's default browser is in the list → it is used
- The user's default browser is not in the list, but another allowed browser is installed → that browser is used instead
- No allowed browser is installed → `a0.browser_not_available` error is returned

On iOS the option is ignored.

## Usage

```typescript
await auth0.webAuth.authorize(parameters, {
  allowedBrowserPackages: [
    'com.android.chrome',
    'com.chrome.beta',
    'com.microsoft.emmx', // Edge
    'com.brave.browser',
    'com.sec.android.app.sbrowser', // Samsung Internet
  ]
});
```

## Files changed

- `src/types/platform-specific.ts` — adds `allowedBrowserPackages?: string[]` to `NativeAuthorizeOptions` with JSDoc indicating Android-only
- `src/specs/NativeA0Auth0.ts` — adds parameter to TurboModule spec
- `src/platforms/native/bridge/NativeBridgeManager.ts` — passes option through to native layer
- `android/src/main/java/com/auth0/react/A0Auth0Module.kt` — implements `BrowserPicker` via `CustomTabsOptions`
- `android/src/main/oldarch/com/auth0/react/A0Auth0Spec.kt` — updates old arch abstract method signature

## References

- Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1976809
- Equivalent Flutter fix: https://github.com/auth0/auth0-flutter/issues/392
- Auth0.Android BrowserPicker: https://github.com/auth0/Auth0.Android/pull/353


subhankarmaiti:
Hey @mrbrentkelly, thanks for this PR! I've reviewed it against the underlying Auth0.Android SDK and have a few things that need to be addressed before we can merge.  

1. The changes should also needs to be added in logout flow otherwise the logout will still be used in users default browser 
2. Please change the name to `allowedBrowserPackages` to keep it same as all the layers.
3. Please add supported test cases
4. Looks like iOS bridge might going to crash at runtime as the changes added a new parameter in the iOS webauth call

mrbrentkelly:
Hey @subhankarmaiti, I think I've addressed all your comments if you want to take a second pass.

Thanks for calling out the iOS side, I had noticed as well it was crashing so I've updated the appropriate method signatures in `A0Auth0.mm` 👍

mrbrentkelly:
Hi @subhankarmaiti, anything else you need from my side here?

subhankarmaiti:
Hi @mrbrentkelly 
Could you update `EXAMPLES.md` to include the usage of this feature?

mrbrentkelly:
@subhankarmaiti, updated 👍

subhankarmaiti (approved):
LGTM

PR #1517, opened 2026-04-23
feat(mfa): Implement Multi-Factor Authentication (MFA) support

subhankarmaiti:
Adds first-class MFA support to the SDK across iOS, Android, and Web, exposing the Flexible Factors flow through a single consistent API:    
                                                                                                                               
  - `mfa.getAuthenticators()` — list a user's enrolled factors                                                                                 
  - `mfa.enroll()` — enroll a new factor (OTP, SMS, voice, email, push)                                                        
  - `mfa.challenge()` — trigger a challenge for an existing factor                                                                             
  - `mfa.verify()` — verify a code and exchange the MFA token for credentials 
 
An `MfaFactorType` enum and a dedicated `MfaError` (with `MfaErrorCodes`) are exported publicly for type-safe usage and error handling.
                                                                                                                                               
  #### Usage                                                                                                                                   
                                                    
  ```ts                                                                                                                                        
  import { useAuth0, MfaFactorType } from 'react-native-auth0';                                                       
                                                    
  const { mfa } = useAuth0();

  // A login attempt against an MFA-enabled connection returns an mfa_token:
  //   const mfaToken = error.json.mfa_token;                                                                                                  
   
  // List the user's enrolled factors                                                                                                          
  const authenticators = await mfa.getAuthenticators({ mfaToken });                                                   
                                                                                                                                               
  // Verify an existing factor (e.g. TOTP from an authenticator app)
  const credentials = await mfa.verify({ mfaToken, otp: '123456' });                                                                           
                                                                                                                      
  // ...or enroll a new factor                      
  const challenge = await mfa.enroll({                                                                                                         
    mfaToken,
    factorType: MfaFactorType.SMS,                                                                                                             
    phoneNumber: '+12025550135',                                                                                      
  });                                                                                                                                          
   
  if (challenge.type === 'oob') {                                                                                                              
    // user receives the code out-of-band, then:                                                                      
    const credentials = await mfa.verify({ mfaToken, oobCode: '000000' });
  }
  ```

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.kt
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.kt
@@ -26,12 +26,14 @@ import com.auth0.android.result.Credentials
 import com.auth0.android.result.PasskeyChallenge
 import com.auth0.android.result.PasskeyRegistrationChallenge
 import com.facebook.react.bridge.ActivityEventListener
+import com.facebook.react.bridge.Arguments
 import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.bridge.ReactMethod
 import com.facebook.react.bridge.ReadableArray
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.bridge.UiThreadUtil
+import com.facebook.react.bridge.WritableNativeArray
 import com.facebook.react.bridge.WritableNativeMap
 import com.google.gson.Gson
 import java.net.MalformedURLException
@@ -108,6 +110,7 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
     private var useDPoP: Boolean = true
 
     private var auth0: Auth0? = null
+    private var mfaClient: MfaClient? = null
     private var myAccount: MyAccount? = null
     private var passwordless: Passwordless? = null
     private lateinit var secureCredentialsManager: SecureCredentialsManager
@@ -281,6 +284,7 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
 
         this.useDPoP = useDPoP ?: true
         auth0 = Auth0.getInstance(clientId, domain)
+        mfaClient = MfaClient(auth0!!, this.useDPoP, reactContext)
         myAccount = MyAccount(auth0!!, this.useDPoP, reactContext)
         passwordless = Passwordless(auth0!!, this.useDPoP, reactContext)
 
@@ -645,6 +649,30 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
                 handleError(error, promise)
             }
         })
+    }
+
+    @ReactMethod
+    override fun getMfaAuthenticators(mfaToken: String, factorsAllowed: ReadableArray?, promise: Promise) {
+        mfaClient?.getAuthenticators(mfaToken, factorsAllowed, promise)
+            ?: promise.reject("NOT_INITIALIZED", "Auth0 not initialized")
+    }
+
+    @ReactMethod
+    override fun mfaEnroll(mfaToken: String, type: String, value: String?, promise: Promise) {
+        mfaClient?.enroll(mfaToken, type, value, promise)
+            ?: promise.reject("NOT_INITIALIZED", "Auth0 not initialized")
+    }
+
+    @ReactMethod
+    override fun mfaChallenge(mfaToken: String, authenticatorId: String, promise: Promise) {
+        mfaClient?.challenge(mfaToken, authenticatorId, promise)
+            ?: promise.reject("NOT_INITIALIZED", "Auth0 not initialized")
+    }
+
+    @ReactMethod
+    override fun mfaVerify(mfaToken: String, type: String, code: String, bindingCode: String?, scope: String?, audience: String?, promise: Promise) {
+        mfaClient?.verify(mfaToken, type, code, bindingCode, scope, audience, promise)
+            ?: promise.reject("NOT_INITIALIZED", "Auth0 not initialized")
     }
 
     @ReactMethod
--- a/android/src/main/java/com/auth0/react/MfaClient.kt
+++ b/android/src/main/java/com/auth0/react/MfaClient.kt
@@ -1,0 +1,279 @@
+package com.auth0.react
+
+import com.auth0.android.Auth0
+import com.auth0.android.authentication.AuthenticationAPIClient
+import com.auth0.android.authentication.mfa.MfaEnrollmentType
+import com.auth0.android.authentication.mfa.MfaException
+import com.auth0.android.authentication.mfa.MfaVerificationType
+import com.auth0.android.result.Authenticator
+import com.auth0.android.result.Credentials
+import com.auth0.android.result.EnrollmentChallenge
+import com.auth0.android.result.OobEnrollmentChallenge
+import com.auth0.android.result.RecoveryCodeEnrollmentChallenge
+import com.auth0.android.result.TotpEnrollmentChallenge
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReadableArray
+import com.facebook.react.bridge.WritableNativeArray
+import com.facebook.react.bridge.WritableNativeMap
+
+enum class MfaFactorType(val value: String) {
+    PHONE("phone"),
+    EMAIL("email"),
+    OTP("otp"),
+    PUSH("push"),
+    VOICE("voice");
+
+    companion object {
+        fun fromString(type: String): MfaFactorType? =
+            entries.find { it.value == type }
+    }
+}
+
+enum class MfaVerifyType(val value: String) {
+    OTP("otp"),
+    OOB("oob"),
+    RECOVERY_CODE("recoveryCode");
+
+    companion object {
+        fun fromString(type: String): MfaVerifyType? =
+            entries.find { it.value == type }
+    }
+}
+
+class MfaClient(
+    private val auth0: Auth0,
+    private val useDPoP: Boolean,
+    private val reactContext: ReactApplicationContext
+) {
+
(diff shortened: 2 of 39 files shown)


android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -507,6 +519,167 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
         })
     }
 
+    @ReactMethod
+    override fun mfaGetAuthenticators(mfaToken: String, factorsAllowed: ReadableArray?, promise: Promise) {
```
pmathew92: Rename to `getMfaAuthenticators ` ?

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -177,6 +181,7 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
         
         this.useDPoP = useDPoP ?: true
```
pmathew92, line 285: completely unrelated but shouldn't the default behaviour for Dpop be false ?

android/src/main/java/com/auth0/react/A0Auth0Module.kt
pmathew92: `mfaGetAuthenticators` Shouldn't this also be renamed to `getMfaAuthenticators`

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -507,6 +512,30 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
...
+    override fun mfaChallenge(mfaToken: String, authenticatorId: String, promise: Promise) {
+        mfaBridge?.challenge(mfaToken, authenticatorId, promise)
+            ?: promise.reject("NOT_INITIALIZED", "Auth0 not initialized")
+    }
+
+    @ReactMethod
+    override fun mfaVerify(mfaToken: String, type: String, code: String, bindingCode: String?, promise: Promise) {
+        mfaBridge?.verify(mfaToken, type, code, bindingCode, promise)
+            ?: promise.reject("NOT_INITIALIZED", "Auth0 not initialized")
+    }
```
pmathew92, line 676: Noted you are following a pattern of `mfa` prefix  . Ensure this is inline with what the auth0/react SDK also follows to maintain consistency ?

android/src/main/java/com/auth0/react/MfaBridge.kt
```diff
@@ -0,0 +1,179 @@
...
+    private fun createAuthClient(): AuthenticationAPIClient {
+        val client = AuthenticationAPIClient(auth0)
+        if (useDPoP) {
+            client.useDPoP(reactContext)
+        }
+        return client
+    }
+
+    fun getAuthenticators(mfaToken: String, factorsAllowed: ReadableArray?, promise: Promise) {
+        val mfaClient = createAuthClient().mfaClient(mfaToken)
```
pmathew92: Can't we just create one instance on `AuthenticationAPIClient` for this entire class ?
```
private val client by lazy { AuthenticationAPIClient(auth0).apply{
            if (useDPoP) {
           useDPoP(reactContext)
           }
      }
}

File: android/src/main/java/com/auth0/react/MfaClient.kt [hunk 5]
```diff
@@ -0,0 +1,179 @@
...
+                }
+
+                override fun onFailure(error: MfaException.MfaListAuthenticatorsException) {
+                    promise.reject(error.getCode(), error.getDescription(), error)
+                }
+            }
+        )
+    }
+
+    fun enroll(mfaToken: String, type: String, value: String?, promise: Promise) {
```
pmathew92, line 144: Define type as enum so that its always limited to the supported factor

example/android/app/build.gradle
```diff
@@ -82,6 +82,7 @@ android {
         applicationId "com.auth0example"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
+        manifestPlaceholders = [auth0Domain: "dev-481o0jxuhhma72l4.us.auth0.com", auth0Scheme: "${applicationId}.auth0"]
```
pmathew92: remove the domain value from here

ios/A0MfaBridge.swift
```diff
@@ -0,0 +1,169 @@
...
+    private let domain: String
+    private let useDPoP: Bool
+
+    init(clientId: String, domain: String, useDPoP: Bool) {
+        self.clientId = clientId
+        self.domain = domain
+        self.useDPoP = useDPoP
+    }
+
+    private func createAuthentication() -> Authentication {
```
pmathew92: Same comment as in Android. Can't we have 1 instance on `Authentication` fot this bridge class

ios/A0MfaClient.swift
```diff
@@ -0,0 +1,169 @@
...
+                    return dict
+                }
+                resolve(list)
+            case .failure(let error):
+                reject(error.code, error.localizedDescription, error)
+            }
+        }
+    }
+
+    func enroll(mfaToken: String, type: String, value: String?, resolve: @escaping RCTPromiseResolveBlock, reject: @escaping RCTPromiseRejectBlock) {
```
pmathew92, line 110: Make type as an enum instead of String

android/src/main/java/com/auth0/react/A0Auth0Module.kt
subhankarmaiti: We have default value as true for RN SDK. I am going to create a separate PR for making it false which going to be a non breaking changes.

android/src/main/java/com/auth0/react/A0Auth0Module.kt
subhankarmaiti: Yes, it matches the same pattern

android/src/main/java/com/auth0/react/MfaClient.kt
subhankarmaiti: At the bridge boundary, the param has to stay String. React Native TurboModule codegen marshals primitives across the JS↔native boundary. 
For consumers, the real type-safety lives in the TS layer, which exposes a public MfaFactorType

ios/A0MfaClient.swift
subhankarmaiti: For consumer side it already exposing the enum `MfaFactorType`. which throw compile time error if passed any other strings.

android/src/main/java/com/auth0/react/MfaClient.kt
```diff
@@ -0,0 +1,279 @@
...
+        mfaClient.getAuthenticators(factors).start(
+            object : com.auth0.android.callback.Callback<List<Authenticator>, MfaException.MfaListAuthenticatorsException> {
+                override fun onSuccess(result: List<Authenticator>) {
+                    val array = WritableNativeArray()
+                    for (authenticator in result) {
+                        if (narrowPhoneChannels && authenticator.type == "phone") {
+                            val channel = authenticator.oobChannel
+                            if (channel == null || !wantedPhoneChannels.contains(channel)) {
+                                continue
+                            }
```
pmathew92, line 122: @utkrishtsahu  Please validate these logic with the MFA change recently did on Android

pmathew92 (dismissed):

android/src/main/java/com/auth0/react/MfaClient.kt
```diff
@@ -0,0 +1,279 @@
...
+import com.facebook.react.bridge.ReadableArray
+import com.facebook.react.bridge.WritableNativeArray
+import com.facebook.react.bridge.WritableNativeMap
+
+enum class MfaFactorType(val value: String) {
+    PHONE("phone"),
+    EMAIL("email"),
+    OTP("otp"),
+    PUSH("push"),
+    VOICE("voice");
```
NandanPrabhu, line 25: recovery code also should be there

NandanPrabhu (changes requested):

android/src/main/java/com/auth0/react/MfaClient.kt
```diff
@@ -0,0 +1,279 @@
...
+    // the `Authenticator.type` tokens Auth0.Android 3.21.0 filters against with
+    // exact, case-sensitive equality. Note: sms and voice both surface as the
+    // `phone` type, so neither can be isolated here (narrowed later by oobChannel);
+    // otp maps to both `otp` and `totp`.
+    private fun mapFactorType(factor: String): List<String> {
+        return when (factor.lowercase()) {
+            "otp" -> listOf("otp", "totp")
+            "sms", "voice" -> listOf("phone")
+            "email" -> listOf("email")
+            "push" -> listOf("push-notification")
```
NandanPrabhu, line 69: Should we consider recovery code too here?

android/src/main/java/com/auth0/react/MfaClient.kt
```diff
@@ -0,0 +1,279 @@
...
+        if (requested.contains("sms")) wantedPhoneChannels.add("sms")
+        if (requested.contains("voice")) wantedPhoneChannels.add("voice")
+        val narrowPhoneChannels = wantedPhoneChannels.size == 1
+
+        mfaClient.getAuthenticators(factors).start(
+            object : com.auth0.android.callback.Callback<List<Authenticator>, MfaException.MfaListAuthenticatorsException> {
+                override fun onSuccess(result: List<Authenticator>) {
+                    val array = WritableNativeArray()
+                    for (authenticator in result) {
+                        if (narrowPhoneChannels && authenticator.type == "phone") {
```
NandanPrabhu, line 118: @subhankarmaiti can we add one or 2 line comment explaining the business logic here

android/src/main/java/com/auth0/react/MfaClient.kt
subhankarmaiti: This enum is used for enroll, and as we can't enroll `recovery-code`, we do not need it here.

NandanPrabhu (approved):

PR #1555, opened 2026-05-31
fix(codegen): inline imported type refs in NativeA0Auth0 spec (#1553)

### LINKED ISSUE #1553: codegenConfig breaks iOS pod install on New Architecture (RN 0.85+) — TSTypeReference in NativeA0Auth0 spec
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

With React Native New Architecture enabled, `pod install` fails during Codegen while processing `RNAuth0Spec` from `react-native-auth0`.

The TurboModule spec uses imported types (e.g. `Promise<Credentials>`) that React Native Codegen does not support. This matches the failure mode discussed in https://github.com/auth0/react-native-auth0/pull/1037, where removing `codegenConfig` was the recommended interop-layer approach until full TurboModule support is ready.

### Environment

- `react-native-auth0`: **5.6.0** (also reproduced on **5.0.1**)
- `react-native`: **0.85.3**
- `expo`: **^56.0.0**
- Node: **20.x**
- Platform: **iOS** (`pod install` / `expo run:ios`)
- New Architecture: **enabled** (Expo SDK 56 default)

### Expected behavior

`pod install` completes and the app builds with New Architecture.

### Actual behavior

Codegen fails and CocoaPods reports an invalid `Podfile`:

```
[Codegen] Processing RNAuth0Spec
[Codegen] UnsupportedTypeAnnotationParserError: Module NativeA0Auth0: TypeScript type annotation 'TSTypeReference' is unsupported in NativeModule specs.
```

The spec references imported types, for example:

```ts
import type { Credentials } from '../types';
// ...
getCredentials(...): Promise<Credentials>;
webAuth(...): Promise<Credentials>;
```

(`node_modules/react-native-auth0` — `lib/typescript/src/specs/NativeA0Auth0.d.ts` or `src/specs/NativeA0Auth0.ts`)

### Workaround (confirmed)

Remove `codegenConfig` from `react-native-auth0`'s `package.json` (via `patch-package`). Auth0 continues to work via the bridge + New Architecture interop layer. After removing it, Codegen skips `RNAuth0Spec` and `pod install` succeeds.

This aligns with the React WG guidance quoted in https://github.com/auth0/react-native-auth0/pull/1037:

> If you have already added a Codegen spec to your library, but the library is not fully converted to TurboModules/Fabric, we recommend that you delete it before proceeding.

### Suggested fix (either is fine)

**Option A (preferred for now):** Remove `codegenConfig` from published `package.json` until the library is fully TurboModule-ready (same as interop approach in #1037).

**Option B:** Keep `codegenConfig` but change `NativeA0Auth0` so all return/param types are Codegen-compatible (inline object types / primitives only — no imported `Credentials` / `ApiCredentials` references).

### Related issues / PRs

- https://github.com/auth0/react-native-auth0/issues/804 — Android `react_codegen_RNAuth0Spec` linking
- https://github.com/auth0/react-native-auth0/issues/879 — New Architecture support
- https://github.com/auth0/react-native-auth0/pull/1037 — Removed `codegenConfig` for v4 interop; v5.x still ships it

### Reproduction

1. Add `react-native-auth0` to an Expo 56 / RN 0.85 app with New Architecture enabled.
2. Run `cd ios && pod install` (or `npx expo run:ios`).

### Additional context

_No response_

### react-native-auth0 version

5.6.0 

### React Native version

0.85.3

### Expo version

^56.0.0

### Platform

Android, iOS

### Platform version(s)

16.4

benjaminkomen:
### Description

Closes #1553.

`react-native-auth0`'s TurboModule spec (`src/specs/NativeA0Auth0.ts`) currently uses imported / aliased type references (`Credentials`, `ApiCredentials`) and indexed-signature object types (`{ [key: string]: string | Int32 | boolean }`). On RN 0.85+ — the React Native version that ships with Expo SDK 56 — Codegen has tightened its grammar and now rejects every `TSTypeReference` to imported / aliased types in NativeModule specs:

```
[Codegen] Processing RNAuth0Spec
[Codegen] UnsupportedTypeAnnotationParserError: Module NativeA0Auth0:
TypeScript type annotation 'TSTypeReference' is unsupported in NativeModule specs.
```

This blocks both iOS `pod install` and the Android `:react-native-auth0:generateCodegenSchemaFromJavaScript` Gradle task on RN 0.85 with the New Architecture — which is the **Expo SDK 56 default** — making the library effectively unusable on the latest RN/Expo for any project that relies on Codegen running.

### Approach — Option B from #1553

The reporter laid out two options:

- **Option A:** drop `codegenConfig` from `package.json` (mirroring #1037).
- **Option B:** keep `codegenConfig`, but make the spec Codegen-friendly by replacing imported / aliased types with inline primitives, `Object`, or inline object literals.

This PR implements **Option B**. Rationale: Option A is a strictly weaker fix — it disables Codegen processing entirely and gives up the road to native TurboModules. Option B keeps `codegenConfig` working today and leaves the door open for a full TurboModule migration later (a follow-up that would tighten `Object` back into proper schemas, but that work depends on Codegen accepting type aliases first).

### Mapping

| Before | After |
|--------|-------|
| `Int32` (locally aliased to `number`) | `number` |
| `Promise<Credentials>` | `Promise<Object>` |
| `Promise<ApiCredentials>` | `Promise<Object>` |
| `Promise<{ ...inline... }>` | `Promise<Object>` |
| `{ [key: string]: string \| Int32 \| boolean }` | `Object` |
| `{ [key: string]: string \| Int32 }` | `Object` |
| `{ [key: string]: string }` | `Object` |

The `Credentials` / `ApiCredentials` / `LocalAuthenticationOptions` types are still exported from `src/types` and used by the public JS API (CredentialsManager, WebAuth wrappers, etc.). **Only the bridge spec is inlined** — the user-facing surface is unchanged.

### No runtime change

The native iOS (`A0Auth0.mm` / `NativeBridge.swift`) and Android (`A0Auth0Module.kt`) implementations already:

- coerce numeric arguments through `NSInteger` / `Double` / `Int`, so the JS-side widening from `Int32` to `number` doesn't change what reaches native code,
- treat the option dictionaries (`localAuthenticationOptions`, `saveCredentials`, `additionalParameters`, `parameters`, `headers`, `userMetadata`) as untyped `NSDictionary` / `ReadableMap`, so collapsing them to `Object` matches what was already being read.

### Validation

Validated downstream on an Expo SDK 56 / RN 0.85.3 app:

- with this patch applied via `patch-package`, iOS Codegen + `pod install` succeed,
- Android `:react-native-auth0:generateCodegenSchemaFromJavaScript` and `assembleDebug` succeed,
- the app launches and hits Auth0 successfully on a physical iPhone (no `EXC_BAD_ACCESS`, login flow works end-to-end).

The same patch text is also posted as a `patch-package` snippet in #1553 so other affected users can apply it locally while this PR is being reviewed.

### Notes for reviewers

- I considered tightening to inline `{ idToken: string; accessToken: string; ... }` shapes for the credential return types (instead of `Object`). RN's Codegen accepts those, but it would re-introduce `TSTypeReference`-style schemas in the generated headers and the native side already returns plain `NSDictionary`s — so the inline shapes would only buy compile-time strictness without unlocking real TurboModule semantics. Happy to switch to that if the maintainers prefer.
- Alternative Option A (drop `codegenConfig`) is a one-line change and would also close #1553 — let me know if you'd rather take that path; I'm happy to repurpose the branch.

### DIFF
--- a/src/specs/NativeA0Auth0.ts
+++ b/src/specs/NativeA0Auth0.ts
@@ -1,7 +1,20 @@
 import { TurboModuleRegistry, type TurboModule } from 'react-native';
-import type { ApiCredentials, Credentials } from '../types';
 
-type Int32 = number;
+// NOTE: This spec uses only Codegen-compatible inline types (primitives, `Object`,
+// inline object literals). It must not import / reference type aliases such as
+// `Credentials`, `ApiCredentials`, or `Int32`, because RN 0.85+ Codegen rejects
+// any `TSTypeReference` to imported / aliased types in NativeModule specs:
+//
+//   UnsupportedTypeAnnotationParserError: Module NativeA0Auth0:
+//   TypeScript type annotation 'TSTypeReference' is unsupported in NativeModule specs.
+//
+// The native iOS / Android implementations already coerce numbers to NSInteger /
+// Double and treat option dictionaries as untyped maps, so this is a type-level
+// change only — no runtime behavior shifts. The `Credentials` / `ApiCredentials`
+// types remain part of the public JS API surface and are used by the
+// CredentialsManager / WebAuth wrappers that consume this bridge.
+//
+// See https://github.com/auth0/react-native-auth0/issues/1553 for context.
 export interface Spec extends TurboModule {
   /**
    * Get the bundle identifier
@@ -22,34 +35,30 @@ export interface Spec extends TurboModule {
   initializeAuth0WithConfiguration(
     clientId: string,
     domain: string,
-    localAuthenticationOptions:
-      | { [key: string]: string | Int32 | boolean }
-      | undefined,
+    localAuthenticationOptions: Object | undefined,
     useDPoP: boolean | undefined,
-    maxRetries: Int32
+    maxRetries: number
   ): Promise<void>;
 
   /**
    * Save credentials
    */
-  saveCredentials(credentials: {
-    [key: string]: string | Int32;
-  }): Promise<void>;
+  saveCredentials(credentials: Object): Promise<void>;
 
   /**
    * Get credentials with the given scope
    */
   getCredentials(
     scope: string | undefined,
-    minTTL: Int32,
+    minTTL: number,
     parameters: Object,
     forceRefresh: boolean
-  ): Promise<Credentials>;
+  ): Promise<Object>;
 
   /**
    * Check if there are valid credentials
    */
-  hasValidCredentials(minTTL: Int32): Promise<boolean>;
+  hasValidCredentials(minTTL: number): Promise<boolean>;
 
   /**
    * Clear credentials
@@ -62,9 +71,9 @@ export interface Spec extends TurboModule {
   getApiCredentials(
     audience: string,
     scope: string | undefined,
-    minTTL: Int32,
+    minTTL: number,
     parameters: Object
-  ): Promise<ApiCredentials>;
+  ): Promise<Object>;
 
   /**
    * Clear API credentials for a specific audience. Optionally filter by scope.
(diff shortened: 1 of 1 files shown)


benjaminkomen:
Might not be an issue after all, closing.

PR #1566, opened 2026-06-17
feat: Add resumeSession() to recover Web Auth logins after Android process death

### LINKED ISSUE #1541: Android: expose WebAuthProvider.addCallback for process-death recovery
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

## Summary
The underlying `com.auth0.android:auth0` SDK exposes a static
`WebAuthProvider.addCallback(Callback<Credentials, AuthenticationException>)` API. When
the Android OS kills the host process while Chrome Custom Tabs is open, the SDK can
still complete the PKCE token exchange on process restart and notify registered
callbacks. `react-native-auth0` does not expose this callback to JavaScript, so React
Native apps have no way to recover credentials from a post-process-death login without
bridging the native API themselves.
## Impact
On devices with aggressive memory management (common on Samsung One UI), this affects
a meaningful share of users who attempt login. The user completes their credentials in
the browser, the SDK finishes the token exchange, but the JS layer is unaware — the app
re-renders at the login screen.

## What we'd like
A JS-accessible method — something like:

```ts
auth0.webAuth.resumeSession(): Promise<{
  success: true;
  credentials: Credentials;
} | {
  success: false;
  errorCode: string;
} | null>;
```

Called once on cold start (gated on an in-flight attempt marker), it drains any pending post-kill result from the SDK callback and returns the credentials or error to JS.

### Alternatives and current workarounds

_No response_

### Additional context

_No response_

### LINKED ISSUE #1549: Android process death during Universal Login causes failed logins (app killed while user is in Auth0 browser)
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

When a user starts the Auth0 Universal Login flow on Android, the OS aggressively kills the app in the background while the user is in the Auth0 browser completing authentication. When the user finishes and returns to the app, the app has been fully restarted and the SDK has lost its state. the login cannot complete and the user is sent back to the login screen with no error. We are measuring approximately 1,000 failed logins per day from this scenario across various Android devices (Samsung, Nothing Phone, and others).

This has been acknowledged by the Auth0 Android SDK team on the native repo: [#972](https://github.com/auth0/Auth0.Android/issues/972). @pmathew92 confirmed the fix is being included in v4 of the Android SDK. Since react-native-auth0 currently bundles an older version of the native Android SDK, this fix is not yet available to React Native users.

### Reproduction

1. Install the app on an Android device with aggressive battery optimization enabled (Samsung, Xiaomi, Huawei, or any device with "Don't keep activities" enabled in Developer Options as a reliable emulator equivalent).
2. Open the app and tap the login button to start the Auth0 Universal Login flow. The Auth0 browser (Chrome Custom Tab) opens.
3. Before completing the login, switch to another app or press the home button. This puts your app in the background.
4. The OS kills the app process in the background while the user is still in the browser.
5. Return to the browser and complete the login with valid credentials.
6. The browser attempts to deep-link back to the app via the redirect URI. The app restarts from a cold start.
7. Expected: The SDK recovers the OAuth callback and the user is logged in.
8. Actual: The SDK has no record of the in-flight auth attempt. The login silently fails and the user lands on the login screen again.

**Easiest way to reproduce on emulator:**

1. Enable "Don't keep activities" in Developer Options on the emulator.
2. Start the login flow.
3. As soon as the Auth0 browser opens, press the home button.
4. The app is immediately killed (due to the developer option).
5. Complete the login in the browser — the app relaunches but the login does not complete.

### Additional context

_No response_

### react-native-auth0 version

5.4.1

### React Native version

0.81.5

### Expo version

_No response_

### Platform

Android

### Platform version(s)

16

subhankarmaiti:
## Description                                                                                                                   
                                                                                                                                   
  On Android, the OS can kill the app's process while the user is completing login in the browser — common on devices with aggressive memory management (Samsung One UI, Xiaomi MIUI), especially during MFA when the user switches apps to fetch a code. When the browser redirects back, the app cold-starts and the in-flight login is silently lost, dropping the user back on the login screen.                                                                                                             
                                         
  This PR adds `resumeSession()`, which recovers that login. The underlying native SDK finishes the token exchange after the process restarts and buffers the result; calling `resumeSession()` once on cold start drains it and returns the recovered `Credentials` (or `null` if there was nothing to recover).                                                                       
                                                                                                                            
  ```js                                  
  // Hooks
  const { resumeSession } = useAuth0();                                                                                            
  useEffect(() => {
    resumeSession().then((credentials) => {                                                                                        
      if (credentials) {                                                                                                    
        // login interrupted by process death was recovered                                                                        
      }                                                                                                                     
    });                                  
  }, [resumeSession]);
                                                                                                                                   
  // Auth0 class
  const credentials = await auth0.webAuth.resumeSession();                                                                         
  ```                                                                                                                              
                                                                                                                                   
  - Android-only; a no-op resolving `null` on iOS and web, so it is safe to call unconditionally.                                  
  - No native `MainActivity` changes required, so it works the same in bare React Native and Expo.                                 
  - Bumps the bundled Auth0.Android dependency to 3.19.0+, which adds the process-death token-exchange recovery (including DPoP).  
                                                                                                          
  ## References                                                                                                             
                                                                                                                                   
  - Fixes #1541                                                                                                                    
  - Fixes #1549                          
                                                                                                                                   
  ## Checklist                                                                                                              
                                                                                                                                   
  - [x] Unit tests added/updated and passing                                                                                       
  - [x] Docs updated (EXAMPLES.md, FAQ.md)
  - [x] Verified on a real Android device (Home → kill process → complete login → recovered on relaunch) 

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added Android process-death authentication recovery via `resumeSession()`, with lifecycle-aware, main-thread-safe behavior.
  * Exposed session recovery across the web/auth layers; iOS/web resolve to `null` when nothing is recoverable.
* **Documentation**
  * Added “Recovering Login After Process Death (Android)” docs (Hooks + class examples) and expanded the FAQ with the recommended flow.
* **Example App**
  * Updated the Android demo to automatically attempt recovery and show a “Resume Session” action.
* **Bug Fixes**
  * Improved handling when the current activity is unavailable and ensured recovery promises settle safely.
* **Dependencies**
  * Bumped the Android Auth0 SDK to 3.19.0 (and added required AndroidX Credentials versions in the example).
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/android/src/main/java/com/auth0/react/A0Auth0Module.kt
+++ b/android/src/main/java/com/auth0/react/A0Auth0Module.kt
@@ -3,7 +3,10 @@ package com.auth0.react
 import android.app.Activity
 import android.content.Intent
 import android.os.Build
+import android.os.Handler
+import android.os.Looper
 import androidx.fragment.app.FragmentActivity
+import androidx.lifecycle.LifecycleOwner
 import com.auth0.android.Auth0
 import com.auth0.android.authentication.AuthenticationAPIClient
 import com.auth0.android.authentication.AuthenticationException
@@ -38,6 +41,11 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
 
     companion object {
         const val NAME = "A0Auth0"
+
+        // Grace window (ms) for an in-flight restored token exchange to complete before
+        // resumeWebAuthSession resolves null.
+        private const val RESUME_SESSION_GRACE_MS = 1500L
+
         private const val CREDENTIAL_MANAGER_ERROR_CODE = "CREDENTIAL_MANAGER_ERROR"
         private const val BIOMETRICS_AUTHENTICATION_ERROR_CODE = "BIOMETRICS_CONFIGURATION_ERROR"
         
@@ -166,19 +174,72 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
         }
         
         builder.withParameters(cleanedParameters)
-        builder.start(reactContext.currentActivity as Activity,
-            object : com.auth0.android.callback.Callback<Credentials, AuthenticationException> {
+        // start() registers a LifecycleObserver internally (Auth0.Android 3.19.0+ for
+        // process-death recovery), which must happen on the main thread.
+        UiThreadUtil.runOnUiThread {
+            builder.start(reactContext.currentActivity as Activity,
+                object : com.auth0.android.callback.Callback<Credentials, AuthenticationException> {
+                    override fun onSuccess(result: Credentials) {
+                        val map = CredentialsParser.toMap(result)
+                        promise.resolve(map)
+                        webAuthPromise = null
+                    }
+
+                    override fun onFailure(error: AuthenticationException) {
+                        handleError(error, promise)
+                        webAuthPromise = null
+                    }
+                })
+        }
+    }
+
+    @ReactMethod
+    override fun resumeWebAuthSession(promise: Promise) {
+        val activity = reactContext.currentActivity
+        if (activity !is LifecycleOwner) {
+            // No lifecycle owner to register against; nothing to recover.
+            promise.resolve(null)
+            return
+        }
+        val lifecycleOwner = activity as LifecycleOwner
+
+        UiThreadUtil.runOnUiThread {
+            val resolved = java.util.concurrent.atomic.AtomicBoolean(false)
+            val loginCallback = object :
+                com.auth0.android.callback.Callback<Credentials, AuthenticationException> {
                 override fun onSuccess(result: Credentials) {
-                    val map = CredentialsParser.toMap(result)
-                    promise.resolve(map)
-                    webAuthPromise = null
+                    if (resolved.compareAndSet(false, true)) {
+                        promise.resolve(CredentialsParser.toMap(result))
+                    }
                 }
 
                 override fun onFailure(error: AuthenticationException) {
-                    handleError(error, promise)
-                    webAuthPromise = null
+                    if (resolved.compareAndSet(false, true)) {
+                        handleError(error, promise)
+                    }
                 }
-            })
+            }
+            // Logout recovery is out of scope; provide a no-op logout callback.
+            val logoutCallback = object :
+                com.auth0.android.callback.Callback<Void?, AuthenticationException> {
+                override fun onSuccess(result: Void?) {}
+                override fun onFailure(error: AuthenticationException) {}
+            }
+
+            // Registering against an already-RESUMED owner synchronously replays onResume,
+            // draining any result buffered after process-death recovery.
+            WebAuthProvider.registerCallbacks(lifecycleOwner, loginCallback, logoutCallback)
+
+            // Safety net for the rare case where the restored token exchange is still in
+            // flight: give it a short grace window, then resolve null if nothing arrived.
+            if (!resolved.get()) {
+                Handler(Looper.getMainLooper()).postDelayed({
+                    if (resolved.compareAndSet(false, true)) {
+                        promise.resolve(null)
+                    }
+                }, RESUME_SESSION_GRACE_MS)
+            }
+        }
     }
 
     @ReactMethod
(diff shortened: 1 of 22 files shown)


android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -166,19 +174,72 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
         }
         
         builder.withParameters(cleanedParameters)
-        builder.start(reactContext.currentActivity as Activity,
-            object : com.auth0.android.callback.Callback<Credentials, AuthenticationException> {
+        // start() registers a LifecycleObserver internally (Auth0.Android 3.19.0+ for
+        // process-death recovery), which must happen on the main thread.
+        UiThreadUtil.runOnUiThread {
```
NandanPrabhu, line 186: where is UiThreadUtil located? is it auth0.android api or inbuilt api from android SDK

pmathew92, line 186: Whats the need to explicitly put it to uiThread here ? Isn't this callback executed in UI thread ? Would there be any issues if this is not added ?

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -166,19 +174,72 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
...
+            }
+
+            // Registering against an already-RESUMED owner synchronously replays onResume,
+            // draining any result buffered after process-death recovery.
+            WebAuthProvider.registerCallbacks(lifecycleOwner, loginCallback, logoutCallback)
+
+            // Safety net for the rare case where the restored token exchange is still in
+            // flight: give it a short grace window, then resolve null if nothing arrived.
+            if (!resolved.get()) {
+                Handler(Looper.getMainLooper()).postDelayed({
```
pmathew92: This can cause memory leak , if the activity is cleared / destroyed before the `RESUME_SESSION_GRACE_MS` elapses . Ideally we should clear any pending handler callbacks in such scenarios

android/src/main/java/com/auth0/react/A0Auth0Module.kt
subhankarmaiti: its from `com.facebook.react.bridge.UiThreadUtil` a react native built-in library

subhankarmaiti: as `auth0.android` usage `LifecycleObserver`, that can't be run on without main thread, otherwise it throws `IllegalStateException`.

android/src/main/java/com/auth0/react/A0Auth0Module.kt
subhankarmaiti: Updated code to cleared the handler

pmathew92 (approved):

PR #1595, opened 2026-07-15
feat: add actor token support for Custom Token Exchange impersonation & delegation

subhankarmaiti:
This PR adds support for **delegation and impersonation** in Custom Token Exchange flows, following [RFC 8693 OAuth 2.0 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693).                                       
                                                                                                
You can now pass an `actorToken` (and its `actorTokenType`) to `customTokenExchange` to indicate the acting party in a delegation or impersonation scenario. When present, the issued tokens carry the `act` claim, which is now exposed on the user profile.              
                                                                                                  
Supported on **web, iOS, and Android**.                                                         
                                                                                                  
```ts                                                                                         
  const credentials = await auth0.customTokenExchange({
    subjectToken: 'external-subject-token',                                                       
    subjectTokenType: 'urn:acme:legacy-token',
    actorToken: 'acting-party-token',                                                             
    actorTokenType: 'urn:ietf:params:oauth:token-type:id_token',                                
  });                                    

  // The acting-party claim is available on the decoded user profile:
  const user = auth0.auth.userInfo... // or via parseIdToken / useAuth0().user                    
  console.log(user.act); // { sub: '...', act: { ... } } for delegation chains    
```

## Summary by CodeRabbit

* **New Features**
  * Added delegation/impersonation support to RFC 8693 custom token exchange using optional `actorToken`/`actorTokenType` across web, iOS, and Android.
  * Exposed acting party details via the returned user’s `act` claim (including nested delegation).
  * Updated the sample apps to collect and use actor token inputs.
* **Bug Fixes**
  * Added stricter early validation for token type and actor-token parameter pairing.
  * Improved native error handling to preserve specific Auth0 error codes.
* **Documentation**
  * Documented “Delegation & Impersonation (Actor Token)” behavior and constraints.
* **Tests**
  * Added/expanded tests for `act` preservation and actor-token validation/forwarding.


ios/NativeBridge.swift
```diff
@@ -390,20 +390,26 @@ public class NativeBridge: NSObject {
...
         var auth = Auth0.authentication(clientId: self.clientId, domain: self.domain)
         if self.useDPoP {
             auth = auth.useDPoP()
         }
-        
+
         let finalScope = scope ?? "openid profile email"
-        
+
+        var actor: ActorToken?
```
NandanPrabhu: var actorToken: ActionToken?

ios/NativeBridge.swift
```diff
@@ -390,20 +390,26 @@ public class NativeBridge: NSObject {
...
+        if let actorToken = actorToken, let actorTokenType = actorTokenType {
```
NandanPrabhu: if let token = actorToken, let tokentype = actorTokenType {
   actorToken = ActorToken(token: token, tokenType: tokenType)
}

ios/NativeBridge.swift
subhankarmaiti: updated naming

ios/NativeBridge.swift
subhankarmaiti: updated naming

NandanPrabhu (dismissed):

src/platforms/native/bridge/NativeBridgeManager.ts
```diff
@@ -256,7 +258,9 @@ export class NativeBridgeManager implements INativeBridge {
         subjectTokenType,
         audience,
         scope,
-        organization
+        organization,
+        actorToken,
+        actorTokenType
       );
       return new CredentialsModel(credential);
     } catch (e) {
```
kishore7snehil, line 266: The catch here always sets `code` to `custom_token_exchange_failed` in the details, even when the error coming out of `a0_call` is already an `AuthError` carrying the real Auth0 code (for example `invalid_request` when the tenant flag is off, or the actor-token JWT errors). The `name` keeps the original code but `code` gets flattened to the generic string, so `error.name` and `error.code` end up disagreeing, and a developer branching on `code` loses the actual reason.

We need to  surface actor-token failures  cleanly, shall we pass the original code through when the error is already an `AuthError`, and only fall back to the generic one for truly unknown errors?

src/core/utils/validation.ts
```diff
@@ -78,3 +78,31 @@ export function validateParameters(
...
+ *
+ * @param actorToken The actor token, if provided.
+ * @param actorTokenType The actor token type URI, if provided.
+ * @throws {AuthError} If exactly one of the two parameters is provided.
+ */
+export function validateActorTokenParameters(
+  actorToken?: string,
+  actorTokenType?: string
+): void {
+  const hasToken = actorToken !== undefined && actorToken !== null;
```
kishore7snehil: The presence check treats an empty string as provided, so `actorToken: ''` with a real `actorTokenType` passes the pairing check and gets forwarded, and the empty value only fails later at the server. The web adapter then spreads it through as `actor_token: ''` too.

Can we treat empty (and maybe whitespace-only) strings as not provided here, so the guard catches this case client-side like it does for the missing one?

src/core/utils/validation.ts
```diff
@@ -78,3 +78,31 @@ export function validateParameters(
...
+ *
+ * Auth0 requires both `actor_token` and `actor_token_type` to be present when
+ * performing delegation. This surfaces a clear client-side error before the
+ * network request is made rather than relying on the server response.
+ *
+ * @param actorToken The actor token, if provided.
+ * @param actorTokenType The actor token type URI, if provided.
+ * @throws {AuthError} If exactly one of the two parameters is provided.
+ */
+export function validateActorTokenParameters(
```
kishore7snehil, line 125: This validates the pairing but not that `actorTokenType` is a valid URI. The product spec calls for a client-side URI check that throws before the network call, the same way `subjectTokenType` is meant to be validated, so a bad type is rejected locally with a clear error rather than bouncing off the server.

Shall we add the URI check here for `actorTokenType` (and confirm `subjectTokenType` gets the same treatment, since I couldn't find a client-side check for it either)?

src/platforms/web/adapters/__tests__/WebAuth0Client.spec.ts
```diff
@@ -484,6 +484,58 @@ describe('WebAuth0Client', () => {
...
+
+      expect(mockSpaClient.loginWithCustomTokenExchange).toHaveBeenCalledWith(
+        expect.objectContaining({
+          actor_token: 'actor-token',
+          actor_token_type: 'http://corporate-idp/id-token',
+        })
+      );
+    });
+
+    it('should not include actor token keys when not provided', async () => {
```
kishore7snehil, line 521: The refresh-token suppression behavior is documented but nothing pins it in a test. When an actor token is used the server returns no refresh token, and we want to be sure the SDK handles a missing `refresh_token` without breaking.

Could we add a case that runs the exchange with an actor token, has the mock return no `refresh_token`, and asserts `credentials.refreshToken` comes back undefined?

kishore7snehil (changes requested):

EXAMPLES.md
```diff
@@ -1061,6 +1062,45 @@ const credentials = await customTokenExchange({
...
+  subjectTokenType: 'urn:acme:legacy-system-token',
+  actorToken: 'actor-id-token',
+  actorTokenType: 'http://corporate-idp/id-token',
+});
+
+const user = parseIdToken(credentials.idToken);
+console.log(user.act); // The acting party claim
+```
+
+> ⚠️ **Refresh token suppression**: When an actor token is present, Auth0 will **not** issue a refresh token, regardless of whether `offline_access` is in the requested scope. `credentials.refreshToken` will be `undefined` in this case.
```
kishore7snehil: Since there's no refresh token in this flow, it's worth adding what that means downstream: a later refresh-token grant can't re-issue the tokens, so the `act` claim can't be re-emitted either and the acting party is effectively fixed at exchange time. That consequence is the part developers tend to get surprised by, more than the missing token itself.

EXAMPLES.md
```diff
@@ -1061,6 +1062,45 @@ const credentials = await customTokenExchange({
...
+  actorToken: 'actor-id-token',
+  actorTokenType: 'http://corporate-idp/id-token',
+});
+```
+
+`actorTokenType` follows the same URI validation rules as `subjectTokenType` and accepts any developer-defined URI.
+
+> ℹ️ **Prerequisites**: This flow requires the `cte_actor_token` feature flag enabled on your tenant and an [Auth0 Action](https://auth0.com/docs/customize/actions) that calls `api.authentication.setActor()` to set the `act` claim on the issued tokens.
+
+**Accessing the `act` claim**
```
kishore7snehil, line 1082: The example reads `act` off the ID token via `parseIdToken`, which is right, but a line on why would help. Auth0 writes the same `act` claim onto the access token too, but the access token can be opaque, so the ID token is the reliable place to read it. Otherwise someone may reach for the access token and find nothing there.

EXAMPLES.md
```diff
@@ -1061,6 +1062,45 @@ const credentials = await customTokenExchange({
...
+  subjectToken: 'subject-provider-token',
+  subjectTokenType: 'urn:acme:legacy-system-token',
+  actorToken: 'actor-id-token',
+  actorTokenType: 'http://corporate-idp/id-token',
+});
+```
+
+`actorTokenType` follows the same URI validation rules as `subjectTokenType` and accepts any developer-defined URI.
+
+> ℹ️ **Prerequisites**: This flow requires the `cte_actor_token` feature flag enabled on your tenant and an [Auth0 Action](https://auth0.com/docs/customize/actions) that calls `api.authentication.setActor()` to set the `act` claim on the issued tokens.
```
kishore7snehil: The prerequisites cover the flag and the Action but not the actor token itself. Auth0 needs it to be a valid JWT and checks the format, signature, expiry, and client_id match, rejecting it otherwise. A short note here would save people a confusing server-side rejection when they pass an opaque or mismatched token.

kishore7snehil (changes requested):

EXAMPLES.md
kishore7snehil: This flag name is internal to us, so we can avoid calling it our explictly.

kishore7snehil (changes requested):

kishore7snehil (approved):
LGTM!

PR #1597, opened 2026-07-16
feat: enforce IPSIE session_expiry with a SESSION_EXPIRED error

subhankarmaiti:
Adds support for the [IPSIE  SL1](https://openid.github.io/ipsie-openid-sl1/draft-openid-ipsie-sl1-profile.html) `session_expiry` claim. When an enterprise connection sets an upstream session ceiling, Auth0 stamps a `session_expiry` timestamp into the ID token, and the underlying platform SDKs enforce it as a hard ceiling on every credential retrieval — past the ceiling the stored credentials are cleared and refresh is skipped, so the user must re-authenticate.

it surfaces this outcome as a single, platform-agnostic `CredentialsManagerError` with `type === 'SESSION_EXPIRED'` across iOS, Android, and Web.

Credentials also expose an optional `sessionExpiresAt` (Unix seconds), surfaced for display only — enforcement stays in the platform SDKs. On iOS and Web it is decoded from the current ID token's `session_expiry` claim; on Android it reflects the ceiling pinned at initial login (the value the credentials manager actually enforces). It is `undefined` when the connection does not emit the claim.


ios/NativeBridge.swift
```diff
@@ -664,14 +665,18 @@ public class NativeBridge: NSObject {
...
             NativeBridge.tokenTypeKey: self.tokenType,
             NativeBridge.idTokenKey: self.idToken,
             NativeBridge.refreshTokenKey: self.refreshToken as Any,
             NativeBridge.expiresAtKey: floor(self.expiresIn.timeIntervalSince1970),
             NativeBridge.scopeKey: self.scope as Any
         ]
+        if let sessionExpiresAt = self.sessionExpiresAt {
+            dict[NativeBridge.sessionExpiresAtKey] = floor(sessionExpiresAt.timeIntervalSince1970)
+        }
+        return dict
```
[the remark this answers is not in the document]
subhankarmaiti: `sessionExpiresAt` is a **derived** value, not stored state. Auth0.swift computes the session ceiling from the `session_expiry` claim inside the stored **ID token** on every read — it is not a persisted `Credentials` slot, and the `Credentials` initializer doesn't accept one. Since `saveCredentials` preserves the ID token, the ceiling is fully reconstructed on the next `getCredentials()` and enforcement is unaffected. `asDictionary()` only exposes the field for JS-side display. Adding a stored slot would introduce a redundant field that could drift from the authoritative claim, so there's nothing to persist here. No change.

README.md
```diff
@@ -754,6 +754,9 @@ try {
...
+| `SESSION_EXPIRED`     | `SESSION_EXPIRED`                                                                                                                                                                                                                                                                                                                                | `sessionExpired`    | `session_expired`                                           |
+
+> **`SESSION_EXPIRED`** is raised when the upstream IdP session ceiling (IPSIE `session_expiry` claim) has been reached. The local session is no longer valid and the user must re-authenticate. See [IPSIE Session Expiry](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md#ipsie-session-expiry) for details.
```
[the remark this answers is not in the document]
subhankarmaiti: These catch bindings are pre-existing and consistent across the whole file (the PR only added an `instanceof` rethrow inside them, not the `any` annotation). Converting all catch clauses to `unknown` with error-record narrowing is unrelated refactoring that touches code this PR didn't otherwise modify, against the repo's surgical-change guideline. Happy to do it as a separate lint/cleanup PR if desirable, but it's out of scope here. No change in this PR.

src/platforms/web/adapters/WebCredentialsManager.ts
```diff
@@ -72,6 +102,20 @@ export class WebCredentialsManager implements ICredentialsManager {
         detailedResponse: true,
       });
 
+      // @auth0/auth0-spa-js enforces the IPSIE `session_expiry` ceiling silently:
+      // once the upstream IdP session has expired it resolves without a token
+      // instead of throwing. Surface this as SESSION_EXPIRED so callers get the
+      // same actionable signal as native (re-authentication required).
+      if (!tokenResponse) {
```
kishore7snehil, line 122: The ceiling is surfaced as `SESSION_EXPIRED` here, but unlike `getCredentials` this path doesn't put `sessionExpiresAt` on what it returns, since API credentials don't carry it. That's a reasonable split, API credentials are a per-audience access token rather than the session, but it does mean the two retrieval paths now expose different session info.

Is it intentional?

src/platforms/web/adapters/WebCredentialsManager.ts
```diff
@@ -72,6 +102,20 @@ export class WebCredentialsManager implements ICredentialsManager {
         detailedResponse: true,
       });
 
+      // @auth0/auth0-spa-js enforces the IPSIE `session_expiry` ceiling silently:
```
kishore7snehil: This ceiling block and the `instanceof CredentialsManagerError` rethrow just below are copied verbatim from `getCredentials`. Two call sites isn't much, but if the ceiling handling ever changes it'll need editing in both places.

kishore7snehil (changes requested):

src/platforms/web/adapters/WebCredentialsManager.ts
```diff
@@ -31,6 +31,20 @@ export class WebCredentialsManager implements ICredentialsManager {
...
+      // once the upstream IdP session has expired it resolves without a token
+      // instead of throwing. Surface this as SESSION_EXPIRED so callers get the
+      // same actionable signal as native (re-authentication required).
+      if (!tokenResponse) {
```
kishore7snehil, line 68: This leans on spa-js resolving without a token at the ceiling rather than throwing. That does hold in the pinned version, the silent-renewal path returns undefined once the ceiling is reached, so the branch is live. Since it's an internal behavior of the underlying SDK and not part of its documented surface, a short note pinning it to the version we rely on would help whoever bumps spa-js next revalidate it.

src/platforms/web/adapters/__tests__/WebCredentialsManager.errors.spec.ts
```diff
@@ -70,4 +70,42 @@ describe('WebCredentialsManager Error Handling', () => {
...
+
+  describe('IPSIE session_expiry ceiling', () => {
+    it('should map a silent undefined token response to SESSION_EXPIRED', async () => {
+      // spa-js enforces the session_expiry ceiling silently: past the ceiling
+      // getTokenSilently resolves without a token rather than throwing.
+      (mockSpaClient.getTokenSilently as jest.Mock).mockResolvedValue(
+        undefined
+      );
+
+      await expect(manager.getCredentials()).rejects.toThrow(
```
kishore7snehil: This calls `getCredentials()` twice, once for the `rejects.toThrow` and again inside try/catch to read `err.type`. The type check only runs if the second call also throws, so if that ever stopped throwing the assertion would quietly be skipped. Pulling the rejection from a single call (`rejects.toMatchObject({ type: 'SESSION_EXPIRED' })`) would pin it in one shot.

kishore7snehil (changes requested):

src/platforms/web/adapters/WebCredentialsManager.ts
subhankarmaiti: yes, ApiCredentials is a per-audience token, not the session

src/platforms/web/adapters/WebCredentialsManager.ts
subhankarmaiti: done. `^2.22.0` pin in the helper comment

EXAMPLES.md
```diff
@@ -638,6 +639,58 @@ function MyComponent() {
 2. **Configure adequate overlap period**: Ensure your Auth0 tenant has at least 180 seconds token overlap configured
 3. **Test on real devices**: Simulate network instability during testing to validate retry behavior
 
+## IPSIE Session Expiry
+
+> **Platform Support:** iOS, Android, and Web.
+
+Auth0 supports the [IPSIE SL1](https://openid.github.io/ipsie-openid-sl1/draft-openid-ipsie-sl1-profile.html) `session_expiry` claim, which lets an upstream identity provider (e.g. Okta) set a hard ceiling on how long an Auth0-issued session may live. When your enterprise connection has this option enabled, Auth0 includes a `session_expiry` Unix timestamp in the ID token returned to your app after login.
```
kishore7snehil: This says the value is a Unix timestamp in seconds, but doesn't call out the common mistake of emitting milliseconds. When the ceiling is set by a Post-Login Action, a developer who forgets to divide by 1000 produces a value that reads as tens of thousands of years out, which the platform treats as no ceiling, so enforcement silently switches off rather than erroring.

Shall we add a short warning that the claim is in seconds and a milliseconds value disables the ceiling? You can refer to `auth0-server-js `

EXAMPLES.md
```diff
@@ -638,6 +639,58 @@ function MyComponent() {
...
+
+The underlying platform SDKs enforce this ceiling on every credential retrieval. Once the ceiling has passed, `getCredentials()` clears the stored credentials and rejects instead of attempting a token renewal — the user must re-authenticate. **No opt-in code is required**; enforcement is transparent once the connection option is active on your tenant.
```
kishore7snehil, line 655: This points at "when your enterprise connection has this option enabled" without naming the option. The other guides name it directly, the connection needs  "Use ID Token for Session Expiry" toggle in the Dashboard on an okta or oidc enterprise connection, using the Auth Code flow.

We can add it here so that developer knows exactly what to switch on.

EXAMPLES.md
```diff
@@ -638,6 +639,58 @@ function MyComponent() {
...
+        throw error;
+      }
+    }
+  };
+
+  // ...
+}
+```
+
+If you need to read the ceiling directly — for example to warn the user before their session ends — it is exposed as `sessionExpiresAt` (an absolute Unix timestamp, in seconds) on the returned `Credentials`. It is `undefined` when the connection does not emit the claim:
```
kishore7snehil, line 686: One thing that's clear in the `auth0-server-js` guides but not here: this ceiling sits on top of the existing idle and absolute session timeouts rather than replacing them, so the session ends at whichever limit comes first. A sentence saying that would set expectations for anyone who already reasons about those timeouts.

EXAMPLES.md
```diff
@@ -638,6 +639,58 @@ function MyComponent() {
...
+
+```jsx
+const credentials = await getCredentials();
+if (credentials.sessionExpiresAt) {
+  const endsAt = new Date(credentials.sessionExpiresAt * 1000);
```
kishore7snehil, line 691: Since this shows reading the value for a countdown, it's worth noting that enforcement applies a small leeway (about 30 seconds) for clock skew, so the session is treated as expired slightly before this exact timestamp. Otherwise someone building a countdown off this value may be surprised the session ends a bit early.

kishore7snehil (changes requested):

src/platforms/web/adapters/__tests__/WebCredentialsManager.spec.ts
```diff
@@ -103,6 +103,40 @@ describe('WebCredentialsManager', () => {
...
+    it('should leave sessionExpiresAt undefined when the claim is absent', async () => {
+      mockSpaClient.getTokenSilently.mockResolvedValue(mockTokenResponse);
+      mockSpaClient.getIdTokenClaims.mockResolvedValue(mockIdTokenClaims);
+
+      const result = await credentialsManager.getCredentials();
+
+      expect(result.sessionExpiresAt).toBeUndefined();
+    });
+
+    it('should reject a millisecond-valued session_expiry (out of range)', async () => {
```
kishore7snehil: This covers the millisecond case, but the decode guard also rejects strings, floats, zero, and negatives, and the exact upper bound isn't pinned either. Those paths aren't exercised, so a future tweak to the guard could loosen one without a test catching it.

Shall we add test for these cases?

kishore7snehil (changes requested):

EXAMPLES.md
subhankarmaiti: Addressed

EXAMPLES.md
subhankarmaiti: Addressed

EXAMPLES.md
subhankarmaiti: Addressed

src/platforms/web/adapters/__tests__/WebCredentialsManager.spec.ts
subhankarmaiti: Addressed

kishore7snehil (approved):
LGTM!

PR #1616, opened 2026-07-30
chore: add build provenance workflow

nirmal-joishi-auth0:
## ✏️ Changes

This pull request adds a security hardening workflow from the unified-pipeline-template (https://github.com/atko-cic/unified-pipeline-template/security). **No functional changes are introduced.**

### Build Provenance

This PR adds `.github/workflows/provenance.yml` from the unified-pipeline-template.

> **⚠️ Before merging, review the added `.github/workflows/provenance.yml` and make the changes described below, plus any other adjustments your CI environment requires.**

> **📦 Publishing to the npm registry?** Do not use `actions/attest-build-provenance` for npm-registry provenance (the npmjs.com badge). Instead, run `npm publish --provenance` (npm ≥ 9.5.0, GitHub-hosted runner, `id-token: write`, and a `repository` field in `package.json` that matches case-sensitively). For a package's first publish, also add `--access public`: `npm publish --provenance --access public`. The npm CLI generates and signs provenance directly to the registry; `attest-build-provenance` only produces a GitHub-side attestation.
>
> **If this repo only builds and publishes an npm package and already runs `npm publish --provenance`, build provenance is fully covered — close and discard this PR.**

> **⚠️ Required before attestation is meaningful — the build must happen in this job:**
> 1. **Build & pass:** replace the `Build artifact` placeholder step with your real build command so the artifact exists on disk before the `Attest build provenance` step runs, then set `subject-path` to that artifact.
> 2. **Adapt into your existing build job:** move the `Attest build provenance` step into the job that already builds and publishes your artifact — provenance must live in the **same job** that produces it. Attesting from a separate job would be a false supply-chain claim.
>
> Optionally, add an `upload-artifact` step after attestation to pass the attested artifact to downstream jobs or release assets.

### Placeholders to fill in before merging

| Placeholder | Description |
|---|---|
| `echo "Replace this step with your build command"` | Replace with your real build command (e.g. `go build`, `mvn package`, `npm run build`) — the artifact must exist on disk before the attestation step |
| `path/to/your/artifact` | Set `subject-path` to the concrete path (or glob) of the built artifact to attest |

## 🔮 Type of Change

- [x] Standard

## 🔗 References

https://auth0team.atlassian.net/browse/SEPIO-1674

- [x] I added at least one link (task, slack thread, etc) to explain why this change is needed.

## 📖 Documentation

No user-facing changes have been introduced.

- [x] I reflected this change in the (internal and/or user-facing) documentation, or added an explanation for why no documentation update is needed.

## 🎯 Testing

This change adds a CI workflow only; validated by the workflow running on this PR.

- [x] This change has integration, unit, or performance test coverage, or I explained why not.

## 🚀 Deployment

- [x] This change can support multiple releases of the code serving traffic at the same time.

## 🔥 Rollback

Reverting this PR removes the added workflow file — no further action required.

- [x] I explained what the rollback for this change will look like.

### DIFF
--- a/.github/workflows/provenance.yml
+++ b/.github/workflows/provenance.yml
@@ -1,0 +1,92 @@
+name: Build with Provenance
+
+# Reference template: build provenance via Sigstore-backed attestation.
+# ─────────────────────────────────────────────────────────────────────
+# Copy the `build` job below into your own workflow. Do NOT copy it into
+# unified-security-pipeline.yml — provenance must live in the same job that
+# builds and publishes the artifact so the GitHub OIDC identity honestly
+# binds to the run that produced it. Attesting from a different job or
+# workflow would be a false supply-chain claim.
+#
+# Order of operations (required)
+# ───────────────────────────────
+#   1. checkout
+#   2. <your build command>  — artifact file must exist on disk before step 3
+#   3. attest-build-provenance — computes artifact digest and signs it
+#   4. (optional) upload / publish the artifact
+#
+# Permissions (caller job must declare all three)
+# ────────────────────────────────────────────────
+#   contents: read      — checkout
+#   id-token: write     — request OIDC token; Sigstore Fulcio uses this to issue
+#                         a short-lived signing certificate bound to this workflow run
+#   attestations: write — store the signed attestation on GitHub
+#
+# Verification (consumer side)
+# ─────────────────────────────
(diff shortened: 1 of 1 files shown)


subhankarmaiti:
closing this pr as it is open for long time, plz reopen when ready

PR #1637, opened 2026-08-13
feat: allow configuring Auth0.Android's native networking client

NandanPrabhu:
## Summary
- Adds `networkingOptions` to `Auth0Options`, exposing Auth0.Android's `DefaultClient.Builder` (connect/read/write/call timeouts, default headers, debug-only request logging) for every native request the SDK makes (web auth, credential renewal, MFA, passkeys, My Account API).
- Android only — accepted and ignored on iOS for API compatibility. `enableLogging` is intentionally debug-only since Auth0.Android logs full request/response bodies (including tokens) at that level.
- Threaded through the core config signature (so the option triggers native re-init when changed), the TurboModule spec, the native bridge manager, and the native adapter.

## Test plan
- [x] `yarn typecheck` — clean
- [x] `yarn test` — 36/36 suites, 712/712 tests pass (new coverage in `configSignature.spec.ts`, `NativeAuth0Client.spec.ts`, `NativeBridgeManager.spec.ts`)
- [x] `yarn lint` — clean on all changed files
- [x] Android unit tests (`./gradlew :react-native-auth0:testDebugUnitTest`) — new `A0Auth0ModuleNetworkingOptionsTest` (2/2) verifies `readTimeout` and `defaultHeaders` are genuinely applied to the built `DefaultClient` against a real `MockWebServer`
- [x] `EXAMPLES.md` updated with a new "Networking Configuration" section


android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -290,6 +310,7 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
 
         this.useDPoP = useDPoP ?: false
         auth0 = Auth0.getInstance(clientId, domain)
+        androidNetworkingOptions?.let { auth0!!.networkingClient = buildNetworkingClient(it) }
```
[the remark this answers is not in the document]
subhankarmaiti: Worth flagging this is broader than re-init: `syncNativeConfig()` re-points the shared singleton before bridge calls, so a sibling client with no `androidNetworkingOptions` silently inherits another client's timeouts and `defaultHeaders`. Resetting to `DefaultClient()` when options are absent fixes both.

src/types/common.ts
```diff
@@ -214,9 +214,48 @@ export interface Auth0Options {
...
+  /**
+   * Enables verbose HTTP request/response logging to Logcat.
+   *
+   * @remarks
+   * **Debug-only.** Auth0.Android logs full request and response bodies at this level,
+   * which includes access, refresh, and ID tokens in plaintext for token-endpoint calls.
+   * Never enable this in production.
+   * @default false
+   */
+  enableLogging?: boolean;
```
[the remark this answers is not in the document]
subhankarmaiti: Agreeing with this: the in-code comment at `A0Auth0Module.kt:71` and the PR description both describe `enableLogging` as debug-only, but nothing enforces it. Gating on `ApplicationInfo.FLAG_DEBUGGABLE` would make the docs true rather than aspirational.

android/build.gradle
```diff
@@ -79,6 +79,9 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation "androidx.browser:browser:1.10.0"
   implementation 'com.auth0.android:auth0:4.0.1'
+
+  testImplementation 'junit:junit:4.13.2'
```
subhankarmaiti, line 83: This is the repo's first Android unit test, but no workflow runs Gradle: `main.yml` only runs `yarn test:ci`, and there's no `gradlew` in `android/` or a `test:android` script. Can we wire it into CI so it doesn't rot?

src/types/common.ts
```diff
@@ -214,9 +214,48 @@ export interface Auth0Options {
    * @remarks Native only (iOS/Android). Has no effect on the web platform.
    */
   credentialsManagerStorageKey?: string;
+  /**
+   * Configures the native networking client (OkHttp) that Auth0.Android uses for every
+   * request it makes (web auth token exchange, credential renewal, MFA, passkeys, etc.).
+   * @remarks Android only. Accepted on iOS for API compatibility but has no effect.
+   */
+  androidNetworkingOptions?: AndroidNetworkingOptions;
```
subhankarmaiti: Every other platform-specific option here documents the restriction in `@remarks` rather than the name: `maxRetries` is iOS-only, `credentialsManagerStorageKey` and `localAuthenticationOptions` are native-only. Since v6 hasn't shipped, would `networkingOptions` be more consistent?

subhankarmaiti (commented):
Nice change, and good call verifying the timeouts against a real MockWebServer rather than just asserting the builder was called. Couple of things inline, plus I've replied on the two existing threads.

src/types/common.ts
NandanPrabhu: we should have network options for iOS as well. Although we have URLSession which can be injected in swift sdk with custom instance of URLSession there by custom network config. From RN directly URLSession cannot be passed but what we can do is use the same contract we have for android for swift too. So, network options would be for iOS the same way it is for android too.

android/build.gradle
NandanPrabhu: added the android test workflow

src/platforms/native/bridge/NativeBridge.ts
```diff
@@ -39,14 +40,16 @@ export interface NativeBridge {
...
   initialize(
     clientId: string,
     domain: string,
     localAuthenticationOptions?: LocalAuthenticationOptions,
     useDPoP?: boolean,
     maxRetries?: number,
-    credentialsManagerStorageKey?: string
+    credentialsManagerStorageKey?: string,
+    networkingOptions?: NetworkingOptions
   ): Promise<void>;
```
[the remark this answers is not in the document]
NandanPrabhu: is it required to add a trailing comma after the last argument? Not getting any build errors

src/types/common.ts
```diff
@@ -214,9 +214,48 @@ export interface Auth0Options {
...
+ */
+export interface NetworkingOptions {
+  /** Connection timeout, in seconds. @default 10 */
+  connectTimeout?: number;
+  /** Read timeout, in seconds. @default 10 */
+  readTimeout?: number;
+  /** Write timeout, in seconds. @default 10 */
+  writeTimeout?: number;
+  /** Overall timeout for the entire call, in seconds. `0` means no timeout. @default 0 */
+  callTimeout?: number;
```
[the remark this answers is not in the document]
NandanPrabhu: could be addressed in another PR

android/src/test/java/com/auth0/react/A0Auth0ModuleNetworkingOptionsTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        val recordedRequest = server.takeRequest()
+        assertTrue(recordedRequest.getHeader("X-Custom-Header") == "custom-value")
+    }
+
+    @Test
+    fun `enableLogging is ignored on a non-debuggable build even when requested`() {
+        server.enqueue(MockResponse().setBody("{}"))
+
+        // SECURITY: If the isDebuggable gate is ever removed, Auth0.Android attaches its
```
subhankarmaiti, line 140: I removed the `isDebuggable &&` guard and all 4 tests still passed, with the request body visible in the test output: Auth0.Android's interceptor writes through `java.util.logging`, not `android.util.Log`, so nothing crashes. Could we assert on captured log records instead?

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -64,6 +66,40 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
...
+
+        // Auth0.getInstance() returns a shared singleton per clientId/domain: a sibling client
+        // (or this same client on re-init) must not inherit another initialization's networking
+        // config, so this always resolves to a fresh DefaultClient() when options are absent
+        // rather than leaving the previous networkingClient in place.
+        internal fun resolveNetworkingClient(
+            networkingOptions: ReadableMap?,
+            isDebuggable: Boolean
+        ): DefaultClient =
+            networkingOptions?.let { buildNetworkingClient(it, isDebuggable) } ?: DefaultClient()
```
subhankarmaiti: `DefaultClient()` hits the deprecated constructor, so the Gradle build warns on this line. `DefaultClient.Builder().build()` is the recommended equivalent and has identical defaults.

```suggestion
            networkingOptions?.let { buildNetworkingClient(it, isDebuggable) } ?: DefaultClient.Builder().build()
```

src/types/common.ts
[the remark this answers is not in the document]
subhankarmaiti: Now that the gate has landed, could the `@remarks` say the option is ignored in release builds rather than only advising against it? Otherwise someone debugging a release build sets it and sees nothing.

src/types/common.ts
subhankarmaiti: Worth noting URLSession has no separate connect or write timeout, so reusing the Android contract as is would silently ignore `connectTimeout` and `writeTimeout` on iOS. Should we expose only the mappable subset there (`readTimeout`, `callTimeout`, `defaultHeaders`, `enableLogging`)?

android/src/test/java/com/auth0/react/A0Auth0ModuleNetworkingOptionsTest.kt
NandanPrabhu: done

android/src/main/java/com/auth0/react/A0Auth0Module.kt
NandanPrabhu: updated the code with the suggestion.

src/types/common.ts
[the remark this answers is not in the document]
subhankarmaiti: This thread got resolved but the wording is unchanged: the `@remarks` here and the warning in `EXAMPLES.md` still only advise against production rather than saying release builds ignore the option. Was that intentional, or worth a follow-up?

README.md
```diff
@@ -1081,4 +1081,4 @@ This project is licensed under the MIT license. See the <a href="https://github.
 [license-image]: https://img.shields.io/npm/l/react-native-auth0.svg?style=flat-square
 [license-url]: #license
 [downloads-image]: https://img.shields.io/npm/dm/react-native-auth0.svg?style=flat-square
-[downloads-url]: https://npmjs.org/package/react-native-auth0
+[downloads-url]: https://npmjs.org/package/react-native-auth0
```
subhankarmaiti: This drops the trailing newline from README, which `prettier --check` flags but `yarn lint` will not catch since it only covers js/ts. Looks unrelated to the change, so probably worth reverting.

subhankarmaiti (dismissed):

src/types/common.ts
NandanPrabhu: yes we should do that

subhankarmaiti (approved):

Issue #1584, opened 2026-06-30
Android: minimizing the Chrome Custom Tab rejects `authorize()` with `USER_CANCELLED` while the browser stays alive (zombie session)


mGallee:
### Checklist

- [x] The issue can be reproduced in the [react-native-auth0 sample app](https://github.com/auth0-samples/auth0-react-native-sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/react-native-auth0#readme), [Examples](https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/react-native-auth0/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/react-native-auth0/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/react-native-auth0/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

On Android, when `authorize()` (`useAuth0`) opens the login in a Chrome Custom Tab and the user minimizes the tab using the toolbar minimize button (the "Minimized Custom Tabs" / picture-in-picture feature, enabled by default since Chrome 122), the app returns to the foreground and the `authorize()` promise immediately rejects with a `WebAuthError` of code `USER_CANCELLED` (`a0.session.user_cancelled`).

The problem is that the Custom Tab is **not closed** — it stays alive in a floating/minimized window (a "zombie" browser). The user can re-open it and complete the Auth0 login in the browser, but the SDK has already rejected the promise and abandoned the transaction, so the eventual redirect is dropped and the credentials never reach the app. The result is an inconsistent state: the app believes the user cancelled and is unauthenticated, while the browser shows a successful (or completable) login.

Compounding this, `cancelWebAuth()` is a no-op on Android (the native `A0Auth0Module` implementation just resolves), so the app has no way to programmatically dismiss the zombie tab after the rejection either.

**Expected behavior** (any one of these would resolve it):

- Minimizing/backgrounding the Custom Tab should not be treated as a hard `USER_CANCELLED` that abandons the transaction; the flow should be resumable when the user returns and completes login, **or**
[...]

PR #1647, opened 2026-08-25
fix: adopt withAuthTab() to fix spurious USER_CANCELLED on Android

NandanPrabhu:
### Changes

Android web authentication now launches through Auth0.Android's **Auth Tab** launch mode (`withAuthTab()`) instead of the plain Chrome Custom Tab, for both **login** and **logout**.

- `android/.../A0Auth0Module.kt`
  - `webAuth()` — login builder now calls `withAuthTab()`
  - `webAuthLogout()` — logout builder now calls `withAuthTab()`

**Why:** On Chrome 122+, tapping the Custom Tab's *minimize* button returned control to the host app while the browser stayed alive. The old lifecycle-based detection misread that resume as a user cancellation and rejected `authorize()` / `clearSession()` with `USER_CANCELLED`, even though the user could still complete the flow. Auth Tab delivers a real `ActivityResult`, so cancellation is reported only on genuine cancel (back button / close), eliminating the spurious error.

This is default-on (no configuration flag) — appropriate for the v6 major. It's a purely internal Android implementation change with no public API impact. Documented in `MIGRATION_GUIDE.md` (§9) and `EXAMPLES.md`.

### References

- Related GitHub issue: #1584

### Testing

Verified on a physical device (Samsung Galaxy A06, Chrome 151) running the example app:

- Runtime logs confirm the Auth Tab path is taken on **login**: `CustomTabsController: Launching URI as Auth Tab. Session available: true`
- Universal Login renders in the Auth Tab surface; the minimize control is no longer exposed (the toolbar has only close + overflow), so the Chrome-122 minimize bug is structurally unreachable.

Manual acceptance checks for reviewers:
- Start `authorize()`, background the app mid-flow, return and complete login → resolves with credentials, no `USER_CANCELLED`.
- Start `authorize()`, tap close (✕) → rejects with `USER_CANCELLED` (genuine cancel still works).
- Repeat for `clearSession()` (logout).

- [x] This change has been tested on the latest version of the platform/language or why not — verified on-device (Galaxy A06, Chrome 151); native behavior isn't covered by the JS unit suite.

### Checklist

- [x] I have read the Auth0 general contribution guidelines

## Summary by CodeRabbit

- **Documentation**
  - Documented Auth Tab as the default Android authentication mode for authorization and session clearing.
  - Clarified that Auth Tab requires Chrome 137 or later and falls back to a standard Custom Tab on older versions.
  - Clarified that ephemeral sessions work only in plain Custom Tabs, not Auth Tabs or Trusted Web Activities.
  - Documented corrected cancellation behavior when minimizing the authentication window.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -2783,7 +2783,7 @@ On Android, web authentication defaults to a Custom Tab, which shows a read-only
 
 > **Platform Support:** Android only. This option is ignored on iOS and web.
 
-> **Note:** Since v6, Android web authentication uses **Auth Tab** by default — a Custom Tab launch mode that delivers proper `ActivityResult` callbacks instead of inferring cancellation from lifecycle events. This fixes the spurious `USER_CANCELLED` error when users tap Chrome's minimize button (Chrome 122+). Auth Tab is the default launch mode for regular Custom Tabs, whereas TWA is opt-in and renders full-screen with no URL bar. Because they use different launch mechanisms, enabling `useTrustedWebActivity: true` makes TWA take precedence and Auth Tab is not used.
+> **Note:** Since v6, Android web authentication uses **Auth Tab** by default — a Custom Tab launch mode that delivers proper `ActivityResult` callbacks instead of inferring cancellation from lifecycle events. This fixes the spurious `USER_CANCELLED` error when users tap Chrome's minimize button (Chrome 122+). Auth Tab requires **Chrome 137 or later**; on older browser versions it automatically falls back to a standard Custom Tab. Auth Tab is the default launch mode for regular Custom Tabs, whereas TWA is opt-in and renders full-screen with no URL bar. Because they use different launch mechanisms, enabling `useTrustedWebActivity: true` makes TWA take precedence and Auth Tab is not used.
 
 ### Required setup
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -2783,7 +2783,7 @@ On Android, web authentication defaults to a Custom Tab, which shows a read-only
 
 > **Platform Support:** Android only. This option is ignored on iOS and web.
 
-> **Note:** Since v6, Android web authentication uses **Auth Tab** by default — a Custom Tab launch mode that delivers proper `ActivityResult` callbacks instead of inferring cancellation from lifecycle events. This fixes the spurious `USER_CANCELLED` error when users tap Chrome's minimize button (Chrome 122+). Auth Tab requires **Chrome 137 or later**; on older browser versions it automatically falls back to a standard Custom Tab. Auth Tab is the default launch mode for regular Custom Tabs, whereas TWA is opt-in and renders full-screen with no URL bar. Because they use different launch mechanisms, enabling `useTrustedWebActivity: true` makes TWA take precedence and Auth Tab is not used.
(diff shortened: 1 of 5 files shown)


EXAMPLES.md
```diff
@@ -2783,6 +2783,8 @@ On Android, web authentication defaults to a Custom Tab, which shows a read-only
 
 > **Platform Support:** Android only. This option is ignored on iOS and web.
 
+> **Note:** Since v6, Android web authentication uses **Auth Tab** by default — a Custom Tab launch mode that delivers proper `ActivityResult` callbacks instead of inferring cancellation from lifecycle events. This fixes the spurious `USER_CANCELLED` error when users tap Chrome's minimize button (Chrome 122+). Trusted Web Activity and Auth Tab are independent; TWA is opt-in and renders full-screen with no URL bar, while Auth Tab is the default launch mode for regular Custom Tabs. Both can be used together by enabling `useTrustedWebActivity: true` — the TWA flow will use Auth Tab internally.
```
pmathew92: Auth Tab works only in the newer versions of the browser . On older version we default to the custom chrome tabs . So some users might still see this issue .

``` Both can be used together by enabling `useTrustedWebActivity: true` — the TWA flow will use Auth Tab internally.```
Was this behaviour verified manually.

note in the Auth0.Android SDK
 * Note: [withAuthTab] and [withTrustedWebActivity] are mutually exclusive. If both are set,
         * TWA takes precedence and Auth Tab will not be used. They rely on different underlying
         * launch mechanisms and cannot be combined.

Comment (NandanPrabhu) [reply EXAMPLES.md]: Good catch. Updated the documentation for case when both TWA and auth tab is enabled

File: android/src/main/java/com/auth0/react/A0Auth0Module.kt [hunk 2]
```diff
@@ -219,6 +219,15 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
...
             if (ephemeralSession == true) { withEphemeralBrowsing() }
+
+            // Auth Tab delivers a real ActivityResult instead of inferring cancellation
+            // from lifecycle, fixing spurious USER_CANCELLED when Chrome minimize is tapped.
+            // Auth Tab requires Chrome 137+; older browsers fall back to a standard Custom Tab.
+            //
+            // Note: [withAuthTab] and [withTrustedWebActivity] are mutually exclusive. If both are set,
+            // TWA takes precedence and Auth Tab will not be used. They rely on different underlying
+            // launch mechanisms and cannot be combined.
+            withAuthTab()
```
subhankarmaiti, line 230: `withAuthTab()` here drops `ephemeralSession`: `CustomTabsOptions.toAuthTabIntentBuilder()` only reads `toolbarColor` and never calls `setEphemeralBrowsingEnabled`, so on browsers with Auth Tab support the session is no longer ephemeral. Can we skip `withAuthTab()` when `ephemeralSession` is true?

src/core/utils/deepEqual.ts
```diff
@@ -2,9 +2,9 @@ export function deepEqual<T>(x: T, y: T): boolean {
   if (x === y) {
     return true;
   } else if (
-    typeof x == 'object' &&
+    typeof x === 'object' &&
```
subhankarmaiti: This file looks unrelated to Auth Tab, and it's a partial fix: line 10's `!=` still trips `eqeqeq`. Worth dropping it from this PR or finishing it?

NandanPrabhu: yes. will drop from this PR

android/src/main/java/com/auth0/react/A0Auth0Module.kt
NandanPrabhu: in auth0.android, ephemeral browsing is supported today as part of default custom tab. It is not supported either by auth tab or Trusted web activity. So, we need to make these mutually exclusive and update the docs

NandanPrabhu: updated the documentation for the same. we dont need code change for the same but documentation for the same highlighting if user pass ephemeral session it wont work with TWA and auth tab

subhankarmaiti: Reopening: since `withAuthTab()` is now unconditional, `ephemeralSession` silently no-ops on the default Android path for everyone, not just TWA users, yet the Ephemeral Sessions docs still say Android honours it on Chrome 136+. `androidx`'s `AuthTabIntent.Builder` has `setEphemeralBrowsingEnabled`, so can we wire it through or skip `withAuthTab()` when `ephemeralSession` is set?

subhankarmaiti (approved):
LGTM. The 4.1.0 bump wires ephemeral through the Auth Tab path, so `ephemeralSession` is honoured on the new default launch mode, and the docs match. Worth one on-device check that `ephemeralSession: true` actually applies through Auth Tab, since CI can't assert it.

PR #1660, opened 2026-09-15
feat: add credentialsManager.clearAll() and default minTtl to 60

subhankarmaiti:
Adds a `clearAll()` method to the credentials manager that wipes the entire underlying secure store, and changes the default `minTtl` on `getCredentials`/`getApiCredentials` from 0 to 60 seconds so tokens about to expire are refreshed proactively. Also fixes ID-token claim validation on the native passkey flows so validation failures surface as typed errors.

**What it exposes:** `credentialsManager.clearAll()` (and `clearAll` from `useAuth0()`), which clears all data in the store — on native this may remove non-Auth0 entries sharing the same store; on web it maps to the single spa-js cache.

```ts
// Wipe every credential from the secure store
await auth0.credentialsManager.clearAll();
```


src/core/interfaces/CredentialsManager.ts
```diff
@@ -50,6 +50,22 @@ export interface CredentialsManager {
...
+   * This delegates to the native SDK's `clearAll()`, which wipes the entire storage
+   * backing the credentials manager (the Keychain service on iOS,
+   * EncryptedSharedPreferences on Android), not just the keys this SDK knows about.
+   * If that store is shared with other data in your app, that data will be removed
+   * too. On web this is equivalent to `clearCredentials()`, since `@auth0/auth0-spa-js`
+   * manages a single cache.
+   *
+   * @returns A promise that resolves when the storage has been cleared.
+   */
+  clearAll(): Promise<void>;
```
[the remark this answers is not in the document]
subhankarmaiti: `clearAll()` is an intentional v6 addition that mirrors the native SDK surface — `CredentialsManager.clearAll()` exists on both Auth0.swift and Auth0.Android. The PR targets `v6-development`

ios/NativeBridge.swift
```diff
@@ -275,7 +275,28 @@ public class NativeBridge: NSObject {
 
         resolve(removed)
     }
-    
+
+    @objc public func clearAll(resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
+        do {
+            try credentialsManager.clearAll()
```
NandanPrabhu, line 282: clearAll does Dpop.clearKeyPair as well in which case below where we are clearing dpop key pair if useDpop is enabled is not required

android/src/main/java/com/auth0/react/A0Auth0Module.kt
```diff
@@ -465,6 +465,23 @@ class A0Auth0Module(private val reactContext: ReactApplicationContext) : A0Auth0
         promise.resolve(true)
     }
 
+    @ReactMethod
+    override fun clearAll(promise: Promise) {
+        secureCredentialsManager.clearAll()
```
NandanPrabhu, line 471: same for android too

NandanPrabhu: @subhankarmaiti

subhankarmaiti: done

NandanPrabhu (approved):

PR #1662, opened 2026-09-16
feat(android): make Auth Tab and TWA opt-in via JS bridge

NandanPrabhu:
## Summary

- `withAuthTab()` and `withTrustedWebActivity()` are no longer enabled unconditionally — both are now opt-in, controlled by new optional flags on `NativeAuthorizeOptions` and `NativeClearSessionOptions`
- Adds `useAuthTab?: boolean` (default `false`) to both options interfaces, wired end-to-end through the TurboModule spec → `NativeBridgeManager` → Android Kotlin → iOS ObjC bridge (iOS accepts the param and ignores it, same as `useTrustedWebActivity`)
- `useTrustedWebActivity` was already opt-in; this change brings `useAuthTab` to parity

**Usage:**
```typescript
// Auth Tab (Android only, Chrome 137+, fixes spurious USER_CANCELLED)
await authorize({}, { useAuthTab: true });
await clearSession({}, { useAuthTab: true });

// TWA (unchanged, still opt-in)
await authorize({}, { useTrustedWebActivity: true });
```

When both flags are set, TWA takes precedence (Auth0.Android behaviour, documented in both the type JSDoc and the Android comments).

## Test plan

- [ ] All 796 unit tests pass (`yarn test`)
- [ ] TypeScript typecheck passes (`yarn typecheck`)
- [ ] New `NativeBridgeManager` tests assert `useAuthTab: true` is forwarded correctly for both `authorize` and `clearSession`
- [ ] Existing tests updated to assert `false` default for the new param
- [ ] Android: verify `withAuthTab()` is **not** called when `useAuthTab` is omitted
- [ ] Android: verify `withAuthTab()` is called when `useAuthTab: true`
- [ ] Android: verify TWA takes precedence when both `useAuthTab: true` and `useTrustedWebActivity: true` are set


subhankarmaiti (approved):

src/platforms/native/bridge/NativeBridgeManager.ts
```diff
@@ -115,7 +115,8 @@ export class NativeBridgeManager implements NativeBridge {
       // //The native layer will check for this and ignore if the value is 99
       parameters.additionalParameters ?? {},
       options.allowedBrowserPackages,
-      options.useTrustedWebActivity ?? false
+      options.useTrustedWebActivity ?? false,
+      options.useAuthTab ?? true
```
subhankarmaiti: why we are setting default value as `true`?

src/platforms/native/bridge/NativeBridgeManager.ts
```diff
@@ -130,7 +131,8 @@ export class NativeBridgeManager implements NativeBridge {
       parameters.federated ?? false,
       parameters.returnToUrl,
       options.allowedBrowserPackages,
-      options.useTrustedWebActivity ?? false
+      options.useTrustedWebActivity ?? false,
+      options.useAuthTab ?? true
```
subhankarmaiti: +1

subhankarmaiti (changes requested):

subhankarmaiti (approved):

pmathew92 (approved):

